PR #27, opened 2017-04-04
State processing value fixes

uhurusurfa:
Fix missing DIRECTION and FILE_NAME fields when logging message state to the database

### DIFF
--- a/Server/src/main/java/org/openas2/processor/receiver/AS2MDNReceiverHandler.java
+++ b/Server/src/main/java/org/openas2/processor/receiver/AS2MDNReceiverHandler.java
@@ -49,8 +49,9 @@ public class AS2MDNReceiverHandler implements NetModuleHandler {
 
 		AS2Message msg = new AS2Message();
 
-        Map<Object, Object> options = new HashMap<Object, Object>();
-		options.put("DIRECTION", "SEND");
+
+		msg.getOptions().put("DIRECTION", "SEND");
+
 		byte[] data = null;
 
 		// Read in the message request, headers, and data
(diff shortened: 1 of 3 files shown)


Server/src/main/java/org/openas2/processor/receiver/AS2MDNReceiverHandler.java
```diff
@@ -49,8 +49,9 @@ public void handle(NetModule owner, Socket s)
 
 		AS2Message msg = new AS2Message();
 
-        Map<Object, Object> options = new HashMap<Object, Object>();
-		options.put("DIRECTION", "SEND");
+
+		msg.getOptions().put("DIRECTION", "SEND");
```
VboxNick: For this purpose it's better to use org.openas2.message.BaseMessage#setOption

Server/src/main/java/org/openas2/processor/receiver/AS2ReceiverHandler.java
```diff
@@ -70,8 +70,9 @@ public void handle(NetModule owner, Socket s) {
 
         byte[] data = null;
         BufferedOutputStream out;
-        Map<Object, Object> options = new HashMap<Object, Object>();
-		options.put("DIRECTION", "RECEIVE");
+
+		msg.getOptions().put("DIRECTION", "RECEIVE");
```
VboxNick: For this purpose it's better to use org.openas2.message.BaseMessage#setOption

uhurusurfa:
@VboxNick  - good catch - have changed ... and I will do the squash next time

VboxNick:
@uhurusurfa Great, thanks. I did squash via GitHub web interface.

PR #32, opened 2017-04-18
Fixed exit command

VboxNick:
H2 server thread was still running event after  org.h2.tools.Server#shutdown that prevented server from shutdown. org.h2.tools.Server#shutdown is internal API from H2. Instead org.h2.tools.Server#stop should be used.

### DIFF
--- a/Server/src/main/java/org/openas2/BaseSession.java
+++ b/Server/src/main/java/org/openas2/BaseSession.java
@@ -1,19 +1,26 @@
 package org.openas2;
 
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 import javax.activation.CommandMap;
 import javax.activation.MailcapCommandMap;
+import javax.annotation.Nullable;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.openas2.cert.CertificateFactory;
 import org.openas2.partner.PartnershipFactory;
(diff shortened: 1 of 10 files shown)


uhurusurfa:
@VboxNick      
The module termination sequence is being invoked 3 times when EXIT command is run and this is shown in the logging in the terminal as 3 separate statements of stopping a module for every  configured module.

One set is caused by the getProcessor().stopActiveModules(); call in the DefaultProcessor class stop() method as the destroy of components invokes the destroy method of the processor component that in turn is responsible for stopping all modules. That call needs to be removed.

Then there is the invoking of shutdown() from the shutdown hook when the program exits. That is also calling  session.getProcessor().stopActiveModules(); 

We cannot stop that unless we store a reference to the hook and remove it when the the program is shutdown through other mechanusms but it is doing unnecessary work since the Default processor simply stops  "active" modules (ie. configured in XML as opposed to actually running). It should check the modules state calling isRunning() and only stop it if it is running. That way at least there will be a clean log showing only one shutdown of each module that is active.

uhurusurfa:
@VboxNick 
EXIT command works great now but CTRL+C does not .... you need to do:
CTRL+C
Press ENTER
CTRL+C

Looks like the StreamCommandProcessor is not being shutdown

uhurusurfa:
Closing as this is no longer necessary due to minor tweaks to the original code change

PR #53, opened 2017-09-28
Release

uhurusurfa:


igwtech:
We figure out the problem with the compression when integrating with Liason's ECS. I'll want to add the entry in the documentation. Should I make a branch for a documentation change?

PR #62, opened 2018-01-20
Allow negative TZ offsets.

cobbr2:
As originally coded, this test will only pass with timezone offsets that begin with `+`.


uhurusurfa:
thanks for the correction

PR #71, opened 2018-03-19
Add Remove HTTP Header folding property for Async MDNs

mikeflisher:
Remove header folding was added as an option in https://github.com/OpenAS2/OpenAs2App/pull/39

This was missed from the Async MDN implementation.

PR adds header folding for async MDNs using the same property: `remove_http_header_folding`

Testing performed with passing through an nginx proxy infront of AS2.


uhurusurfa:
Thanks for the contribution

PR #73, opened 2018-03-21
Fix expected service RC

jsmucr:
OpenAS2 server returns RC 143 on successful shutdown and systemd should expect so.


uhurusurfa:
Thanks for contributing

PR #75, opened 2018-04-04
Stepbystep002

lucguinchard:
Adding the contents of OpenAS2HowTo file in README.md
      Added: "Below is a sample configuration for using Mysql database"

A PDF can be generated from these sources.
It makes the versioning of the documentation easier.

### DIFF
--- a/Server/src/main/java/org/openas2/lib/helper/ICryptoHelper.java
+++ b/Server/src/main/java/org/openas2/lib/helper/ICryptoHelper.java
@@ -13,6 +13,9 @@ import org.openas2.message.AS2Message;
 import org.openas2.message.Message;
 
 public interface ICryptoHelper {
+    static final int JCE_LIMITED_MAX_LENGTH = 128;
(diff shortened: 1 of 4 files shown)


Server/src/main/java/org/openas2/lib/helper/ICryptoHelper.java
```diff
@@ -13,6 +13,9 @@
 import org.openas2.message.Message;
 
 public interface ICryptoHelper {
+    static final int JCE_LIMITED_MAX_LENGTH = 128;
+    static final String JCE_LIMITATION_ERROR = "OpenAS2 needs `Java Cryptography Extension` (JCE) Unlimited Strength Jurisdiction Policy Files!\nPlease verify it is installed in your Java installation. You can download it from Oracle Web site.";
```
DevDef: Why the backquotes?
Would be better like this:
"OpenAS2 needs Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy Files!\nPlease verify it is installed in your Java installation. You can download it from Oracle Web site."

VboxNick: I would:
- replace Oracle with "official web site", since trade marks and other legal things.
- inline JCE_LIMITATION_ERROR, since as a constant it's useless.

PR #77, opened 2018-04-11
Add short UUID random parameter

jsmucr:
This PR allows usage of $rand.shortUUID$ to get new random UUID encoded in Base62. This leads into shortening message IDs and therefore shortening file names and log messages as well.


uhurusurfa:
Thanks for the contribution

PR #80, opened 2018-04-12
Allow custom ID pattern for MDN messages sent to partner

jsmucr:
The current configuration options allow user to create ID patterns that are valid for outgoing files but invalid for outgoing MDNs. An example of this behavior could be:

`as2_message_id_format="$msg.sender.name$-$msg.receiver.name$-$date.yyyyMMddHHmmss$-$msg.attributes.transactionID$"`

Such pattern works for outgoing files: The `transactionID` attribute (a token generated outside of OpenAS2) is taken right from the file name and no other UUID is used to keep things clear.
The outgoing MDNs also have their message ID but pattern above isn't valid here and `"null"` is used instead of the transaction ID. This would lead into duplicit message ID problems on the partner side, even if the `transactionID` was strictly unique.

This commit resolves this (even though uncommon) situation by adding an optional parameter to either global or partnership configuration. The user can set up a pattern that would be used for MDNs only.

### DIFF
--- a/Server/src/main/java/org/openas2/message/AS2MessageMDN.java
+++ b/Server/src/main/java/org/openas2/message/AS2MessageMDN.java
@@ -34,7 +34,7 @@ public class AS2MessageMDN extends BaseMessageMDN {
     @Override
     public String generateMessageID() throws InvalidParameterException
     {
-        return org.openas2.util.AS2Util.generateMessageID(getMessage());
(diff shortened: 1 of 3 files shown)


uhurusurfa:
This is now provided by a separate pull request that is merge3d so closing 
See here for implementation: https://github.com/OpenAS2/OpenAs2App/blob/dev/Server/src/main/java/org/openas2/util/AS2Util.java#L65

PR #83, opened 2018-05-18
Close sockets even on empty/invalid HTTP request

jsmucr:
This commit fixes a bug where a socket stays open if an empty or an invalid request has been received. This is a common situation with various TCP layer health check systems.


uhurusurfa:
Thanks for the fix.

PR #84, opened 2018-05-21
Enclose Message-ID and Original-Message-ID headers in angle brackets

jsmucr:
This commit solves incompatibilities with clients such as Mendelson AS2. In this particular case the message sent to Mendelson contains ID not enclosed in angle brackets. This message, however, can never be acknowledged properly because Mendelson encloses Message-ID and Original-Message-ID values into angle brackets by default (as required by RFC 2822). OpenAS2 then considers the brackets to be part of the original message ID and is unable to pair the MDN with the original message.


jsmucr:
I'm glad to hear that. :-) I myself thought I'd try to simplify the logic but I didn't want the patch to be that complex.

uhurusurfa:
NOTE: The fix will not add and remove angle brackets in a similar fashion to your fix. 
Per this section https://tools.ietf.org/html/rfc4130#section-5.3.3 that adds AS2 specific additional restrictions on Message-Id on top of RFC 2822:

> For maximum backward
   compatibility, when receiving a message, do not check for angle
   brackets.  When creating the Original-Message-Id header in an MDN,
   always use the exact syntax as received on the original message;
   don't strip or add angle brackets.

To conform with this, the system default for the Message-Id template will include the angle brackets by default and can be overridden if desired - it will be up to the deployer of the application whether to remain conformant with the AS2 standard if the choose to modify the Message-Id format that the system generates.

jsmucr:
I agree and have no problem with that. But please keep in mind that not every other client behaves like this. Mendelson AS2 is a good example of that (I wouldn't have to tackle with the problem otherwise). Also be aware of Windows file name restrictions (angle brackets aren't allowed).

uhurusurfa:
Yes - I have tested the solution against Mendelson and works fine.

The filename restrictions for Windows is handled already - see section titled "File System Issues" in the documentation.

jsmucr:
I meant file names of pending info files which are message IDs. This is the problem I had to deal with:

    org.openas2.OpenAS2Exception: Could not retrieve pending info file: /opt/openas2/bin/../config/../data/mdn/pending/info/<OpenAS2-Mendelson-20180521081350-noid-nojob-BSmakqt4VF0o6toY7Oz05A> (No such file or directory)
    	at org.openas2.util.AS2Util.getMetaData(AS2Util.java:740)
    	at org.openas2.util.AS2Util.processMDN(AS2Util.java:635)
    	at org.openas2.processor.receiver.AS2MDNReceiverHandler.handle(AS2MDNReceiverHandler.java:122)
    	at org.openas2.processor.receiver.NetModule$ConnectionThread.run(NetModule.java:202)
    Caused by: java.io.FileNotFoundException: /opt/openas2/bin/../config/../data/mdn/pending/info/<OpenAS2-Mendelson-20180521081350-noid-nojob-BSmakqt4VF0o6toY7Oz05A> (No such file or directory)
    	at java.io.FileInputStream.open0(Native Method)
    	at java.io.FileInputStream.open(FileInputStream.java:195)
    	at java.io.FileInputStream.<init>(FileInputStream.java:138)
    	at org.openas2.util.AS2Util.getMetaData(AS2Util.java:737)
    	... 3 more

It happens if I send a file to Mendelson. The message ID contains no brackets and thus the info file name contains no brackets as well. But as Mendelson obviously does not go with the RFC4130 restrictions, it adds the brackets to the Original-Message-ID. This results in the exception above.

uhurusurfa:
Ok - I will think about making it configurable per partnership.

uhurusurfa:
This is now handled in a separate pull request so closing.

PR #89, opened 2018-06-12
Message id encapsulation

uhurusurfa:
Provide ability for Message-ID to be encapsulated in angle brackets to conform with RFC4130 spec.
Handle possibility of no angle brackets in returned MDN or angle brackets added to Original-Message-ID when sent one was not encapsulated


uhurusurfa:
@jsmucr - this should cover all your concerns raised about Message-ID - let me know if you see any issues please

jsmucr:
Thanks a lot. :-) I'll take a look at that.

PR #100, opened 2018-07-25
Use a threadpool to handle incoming connections.

fwiersVANAD:
Re-using threads can be a lot more efficient than creating new threads,
especially under heavy load. There is room for further improvement: e.g. set an upper limit to the amount of threads in the pool to prevent an overload, switch to using daemon threads so that the server can always be stopped after a timeout.


uhurusurfa:
Thanks for the optimization

PR #101, opened 2018-07-25
Option to disable chunking for specific partners.

fwiersVANAD:
Added partnership option no_chunked_max_size that, when set to a
positive number, will set a "Content-Length" header in the HTTP POST
message transfers and disables chunked transfer coding.

This option should not be needed in production, but can help with testing when faced with other AS2 servers that do not support chunked transfers. To make this a production feature, the implementation should use a temporary file to determine size (Content-Length) and stream from the temporary file to prevent abnormal memory usage.


uhurusurfa:
If you know of specific AS2 software vendors that do not support chunking let me know and I can add to documentation.

fwiersVANAD:
It's a bit obscure I guess: 
Liaison(R) Exchange Commerce Suite(TM)
(formely known under a bunch of other names, e.g. iSoft)

PR #105, opened 2018-07-31
Use a more generic solution for compatability with micalg values.

fwiersVANAD:
The option "rename_digest_to_old_name" has generic support in
BouncyCastle, so use that. Solution similar to:
https://github.com/phax/as2-lib/commit/ed08dd00b6d721ec3e3e7255f642045c9cbee9c3

Fix a couple of unhelpful NullPointerExceptions and remove 2 partnership
options that no longer exist.


uhurusurfa:
Awesome ... thanks for the contrib.

PR #107, opened 2018-08-10
Systemd service uses OPENAS2_PID, startup script must use that too.

fwiersVANAD:
Starting OpenAS2 with openas2.service fails when the PID-file location
is changed because the start-script does not use the OPENAS2_PID
variable set by the service. The start-script used PID_FILE.

Also tweak the start-script example with EXTRA_PARMS for using old
certificates ("org.bouncycastle.asn1.allow_unsafe_integer").


uhurusurfa:
Please change to "else if" and restore original to retain backwards compatibility,

uhurusurfa:
... basically alter the script to support boith possible environment variables

fwiersVANAD:
I added an extra commit which checks if PID_FILE is set and OPENAS2_PID is not set. My bash scripting is not that great, so I tried to keep it simple and tested the changes in a separate test-script.

PR #111, opened 2018-08-22
Attribute enhancer support for all components

uhurusurfa:
Provide ability to use the attribute enhancer feature in all components to simplify configuration strings and config changes.

### DIFF
--- a/Server/src/config/config.xml
+++ b/Server/src/config/config.xml
@@ -1,5 +1,6 @@
 <openas2>
     <properties
+		storageBaseDir="%home%/../data"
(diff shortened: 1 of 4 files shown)


fwiersVANAD:
The idea is good but I cannot get this working. The test fails and when started with a compiled version the directory "$properties.storageBaseDir$" gets created instead of the values it should replace. Did some of the changed files not make it to this pull request?

uhurusurfa:
@fwiersVANAD  You are correct :( ... new pull request will add the missing commit.

PR #121, opened 2018-09-14
Consider non base64 encodings being rather binary

jsmucr:
The previous version presumes that 7-bit, 8-bit or even quoted-printable encodings can be handled using base64 decoders which is obviously not true and it's better to handle these as binary.

### DIFF
--- a/Server/src/main/java/org/openas2/lib/helper/BCCryptoHelper.java
+++ b/Server/src/main/java/org/openas2/lib/helper/BCCryptoHelper.java
@@ -528,7 +528,7 @@ public class BCCryptoHelper implements ICryptoHelper {
 
     protected String getEncoding(String contentTxfrEncoding)
     {
-        // Bouncy castle only deals with binary or base64 so pass base64 for 7bit, 8bit etc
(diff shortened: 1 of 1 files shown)


uhurusurfa:
The preferred fix here would be to set the  HTTP header to the value the user sets in the config.
Try changing theis line inlciding the try/catch lines here https://github.com/OpenAS2/OpenAs2App/blob/ffa863557a963b0a57e6c38d38879415ac004c39/Server/src/main/java/org/openas2/processor/sender/AS2SenderModule.java#L516 to this:
		cte = msg.getPartnership().getAttribute(Partnership.PA_CONTENT_TRANSFER_ENCODING);

jsmucr:
Umm... why? I mean - it does solve that problem of mine, but the function itself seems wrong as well. Why would one consider 7 or 8-bit encoding base64 rather than binary?

uhurusurfa:
The main reason is because that is what I believe the specification probably intended .... it should reflect the encoding of the original file being sent. 
With the change you are proposing, the header will never have anything other than "binary".

PR #147, opened 2019-03-14
Configuration ui

uhurusurfa:


Server/src/main/java/org/openas2/cmd/processor/restapi/AuthenticationFilter.java
```diff
@@ -0,0 +1,120 @@
...
+            String usernameAndPassword = new String(Base64.decode(encodedUserPassword.getBytes()));
+
+            //Split username and password tokens
+            final StringTokenizer tokenizer = new StringTokenizer(usernameAndPassword, ":");
+            final String username = tokenizer.nextToken();
+            final String password = tokenizer.nextToken();
+
+            //Verifying Username and password
+            System.out.println(username);
+            System.out.println(password);
```
VboxNick: This one is security issue. If the password should be logged, please consider a mask for the password. For instance, se*******

Server/src/main/java/org/openas2/cmd/processor/restapi/AuthenticationFilter.java
```diff
@@ -0,0 +1,120 @@
...
+            //Decode username and password
+            String usernameAndPassword = new String(Base64.decode(encodedUserPassword.getBytes()));
+
+            //Split username and password tokens
+            final StringTokenizer tokenizer = new StringTokenizer(usernameAndPassword, ":");
+            final String username = tokenizer.nextToken();
+            final String password = tokenizer.nextToken();
+
+            //Verifying Username and password
+            System.out.println(username);
```
VboxNick: Please use Logger for logging

Server/src/main/java/org/openas2/cmd/processor/RestCommandProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+                }
+            }
+        }
+        return result;
+    }
+
+    @Override
+    public void destroy() throws Exception {
+        super.destroy();
+        server.shutdown();
```
VboxNick: wrapped by try-catch would be good.

Server/src/main/java/org/openas2/cmd/processor/RestCommandProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+        /**
+         *
+         *
+         * return secure;
+         *
+         */
+        // create and start a new instance of grizzly http server
+        // exposing the Jersey application at BASE_URI
+        try {
+            if (baseUri.getScheme().equalsIgnoreCase("https")) {
```
VboxNick: a log statement would be helpful.

VboxNick:
Thanks for the PR. Generally it looks good, but I left some comments. There is one things I would like to ask @uhurusurfa is that for me REST/Http management is kind of extension for OpenAS2 and it might be better to have it as a separate module with dependency on the Server module.

Server/src/main/java/org/openas2/cmd/processor/restapi/AuthenticationFilter.java
uhurusurfa: Agreed - I only raised the PR to see the state of the branch nd perhaps pull some of the owork in but will not be merging this brnach.

Server/src/main/java/org/openas2/cmd/processor/RestCommandProcessor.java
igwtech: Logging added

Server/src/main/java/org/openas2/cmd/processor/RestCommandProcessor.java
igwtech: To what purpose? Logging the vent?

Server/src/main/java/org/openas2/cmd/processor/restapi/AuthenticationFilter.java
uhurusurfa: @igwtech  - You will need to use the logger here and as pointed out, the logging of a password is a security issue so you need to remove that

uhurusurfa:
Hi Javier - thanks for the contribution.
This will require the client app to be of any use to anyone.
To include the client app you will need to create a separate project in the same way that the "Remote" package is set up.
Also, you will not need that plugin to be scheduled so you need to override the "schedule" method and avoid scheduling.

igwtech:
Understood Christopher.
 I'll proceeded to add a base client application. We also need to include
some Unit testing for the API to do regression testing.

Regards, J

uhurusurfa:
@igwtech  - This will also require changing the minimum supported version of Java to 1.8 since Grizzly is not part of 1.7 and not worth complicating the deployment. I am also in the process of implemented CI through travis and unit tests for your addon would be perfect.

igwtech:
I wanted to ask you. What is the architectural vision for OpenAS2??
  What I mean Is that Nick mentioned his concerns with bloating up
dependencies for OpenAS2 and I agree with him. But in order to address his
concern and keep the software improving and flexible we need to design some
kind of plugin system and let the modules depend on OpenAS2 without needing
to be compiled or hosted on the same project.
  The other concern I have is with the DefaultProcessor and how it
interacts with the modules.. Looks like the processor is just a holding
container for modules and the modules know how to communicate with each
other using the explicit calls to Handle on the session object. I would
have expected the DefaultProcessor to implement the "process flow" and the
modules to execute the process actions without dependencies or knowledge of
each other.

-J

VboxNick:
I do appreciate your contribution. Thanks. When I think about AS2 Server the first thing that comes into my mind is an STMP server.
It's something small, simple, robust, efficient, secure and can be run without any intervention for reasonable amount of time.
That being said that event existing CommandsProcessor is redundant, IMHO.
Just think about importance of AS2 in EDI processes - types of documents and their atts, peer connections, load, security requirements,
partner and their certificate lifecycle, etc, then think about effort you need to put into development, testing and support.
Does it make sense?

After 3 years of use of OpenAS2 I can say - no, there is no need for UI or any configuration interface.
Let's consider the following deployment scenario:

                        openas2 instance 1 on port 10081
                      /
  nginx on port 80 --
                      \
                        openas2 instance 2 on port 10082

where nginx is just a reverse proxy, two OpenAS2 docker containers are running behind. The OpenAS2 docker image is self-contained system
with all configurations and certificates inside and requires only one parameter - a password for certificate storage.
The image is immutable and can be easily built and deploy in rolling fashion via Ansible in less then one minute.
For this setup 1Gb RAM more then enough. The same image might run in a cloud. In this scenario a configuration interface would be
Ansible or other automation tool. So you don't configure OpenAS2, you're more building and deploying it.
Everything is under VCS, no losses possible, accountability in place as well, changes are tracked.

This scenario might not suit your needs and requirements. Fair enough. Commands exchange via network socket would be
the best thing I can imaging so classical client-server pattern. Two runtimes are seperate - awesome. Perhaps might be too complicated,
then I would consider composition, where a maven module has a dependency on OpenAs2Server and extends functionality and OpenAS2Server
module knows nothing about UI. The code might be place within the same repo and release at the same time.
Perhaps, @uhurusurfa  wouldn't be happy, more work for him.

Thanks,
Nick

igwtech:
I hear you Nick. I also think of AS2 as a regular internet server but my
take of it is like more like an Apache HTTPd or ProFTPd. In each case, the
server is composed of a main Core and multiple plugins which allows the
user to configure his/her systems how they like it. To be clear, I like
that architecture.

  Most users find configuring AS2 a lot more complex and confusing than
SMTP or HTTP, that's why every AS2 capable system I have used has user
interface (Have you tried Mendenson's AS2? ).

  I feel as well the commands should be abstracted (as well as other
functions like db tracking etc) into plugins and let the OpenAS2 core
handle only the AS2 protocol and plugin management.

Regards, J

uhurusurfa:
@igwtech  @VboxNick  
Javier - the OpenAS2 system is designed as a pluggable system though not perhaps in the way you might design it. 
There are certainly other ways to do the plug and play function but the way it works at the moment does not have any significant negatives and there are other things that need improving before I would consider redesigning the pluggable module system. The one change I will be putting in for the plugin modules is to add an "enabled" attribute on each module so that they can be enabled and disabled using a property thus removing the need to delete or comment them out of the config.xml
In terms of future development, the following items are a priority from my own perspective:
1.  AS2 1.2 compatibility simplifies certificate management considerably.
2. A standalone configuration UI that simpliifies the setup and managment for non-technical users as you pointed out.
3. Move the directory polling module config to the partnership.

igwtech:
phew.. Standalone Configuration UI implemented in Vue.js/Bootstrap4. Builds into HTML5+JS+CSS3. Connects directly to OpenAS2 server via Rest API Url.

PR #149, opened 2019-03-14
Msg tracking db table name customization

DevDef:
Could you tell me if I should lowercase this kind of option argument strings:
[String direction = (String) options.get("DIRECTION");](https://github.com/OpenAS2/OpenAs2App/blob/master/Server/src/main/java/org/openas2/processor/msgtracking/BaseMsgTrackingModule.java#L52)

or even better, change it to:
```java
String direction = (String) options.get(FIELDS.DIRECTION);
```


uhurusurfa:
Go ahead and change all the accesses to fields from the resuitset to use FIELDS.xxx

String direction = (String) options.get(FIELDS.DIRECTION);

DevDef:
I’ve just rebuilt my RPMs and it seems to be OK.
I think that you could merge it now…

uhurusurfa:
Thanks for the contribution 🎉

PR #168, opened 2019-11-04
Package reorganization, code cleanup adjustments to simplify creating…

igwtech:
… extensions for Session, and Modules
 - Made some private methods protected
 - Moved the AS2DirectoryPoller into its own package "producer", extracted interface
 - Created "monitor" package for Healthcheck Module and related and Moved NetModuleHandler and NetException into the "support" package
 - Adjusted main and test configuration files


uhurusurfa:
This change is not backwards compatible - it will break every OpenAS2 deployment out there and require eveyone to modify the path in order to upgrade.

If the change is key to making code simpler and easier to add new modules going forward then we will need an upgrade script that is idempotent to automatically change deployments by running a script when upgrading.
So otherwise leave the DirectoryPollingModule where it is and add anything new into the "producer" package.
You can add an automated change to this file and make sure it is idempotent if you feel there is a good reason to move the DirectoryPollingModule:
Server/src/.bin/config.xslt

Server/src/main/java/org/openas2/processor/producer/ProducerModule.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
```
VboxNick: default template. Please remove it.

Server/src/main/java/org/openas2/processor/producer/ProducerModule.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+package org.openas2.processor.producer;
+
+import org.openas2.message.AS2Message;
+
+/**
+ *
+ * @author javier
+ */
+public interface ProducerModule {
```
VboxNick: Please, provide some docs for that.

VboxNick:
1) The formatting seems to inconsistent across the project. Sometimes it is C-style and sometimes it is Java-style. My suggestion is to create a basic check style config and format the whole project in a one shot without any refactoring, changes, etc. In future it will be enforce by check-style and nobody can format something differently so easier to review. I can contribute to that. https://checkstyle.org/

2) Added some comments.

3) XSLT is good thing for such things. However, I would introduce a version for xml-configs and check compatibility during startup. Something like:
```
<?xml version="1.0" encoding="UTF-8"?>
<openas2 config-version="1.0.0">
   ....
</openas2>
```
 If it needs some manual interventions/upgrades, then throw an exception with meaningful message, e.g. with a link to a "Upgrade guide" and stop the app.  Perhaps, the final artifact version might be used there, e.g. 2.9.0. Not sure. The major version of the final artifact (jar) should be increased as well.

4) Please, do atomic commits with some explanations. It's good practice for remote collaboration. You might even create an issue for yourself. 

Thanks for your contribution.

igwtech:
Thanks for the review comments. 
-  I like the code formatting suggestion. I really would like to see consistent formatting across the code.
- Regarding the package change. I was afraid it would break backwards compatibility. We don't need to move the modules of packages, it was more a code maintenance improvement by having the modules organized in a structure that at least in my opinion makes more sense according to the pipeline.
   The only "real" requirements for extensibility would be the changes from private to protected methods and the interface extraction.
  There are several options we can take, an upgrade script like you mention, proxy classes marked as deprecated or configuration versions with differences handled thru code. 
   I would like us to discuss them and decide. 
- As far as the single commit. I'm new to sharing PR publicly on Github. I'm used to do multiple commits on my local repo, is there a way to merge them on a single commit before sharing?

uhurusurfa:
1. Standard code style is fine. If  Javier has not done extensive work with the existing code base then it can be done now and all new dev done on the reformatted version.
2. 
3. I am happy to have a config verwion in the XML. It must be automatically updated with the release script and can be done in the section of the pom.xml that sets the release version in the MANIFEST. 
4. I have no problem with multiple commits locally. It makes it easier to see what was done with each commit. The PR is set up to allow a squash and merge so that in the master branch it appears as an atomic commit.

uhurusurfa:
I've thought a bit about the nomenclature and the DirectoryPollingModule is a receiver since it receives files.
The RabbitMQ is a message Q that can both receive and send files so rather than a producer only, it is both a producer and consumer so it could be placed in a **transceiver** package.
The module code that reads from the RabbitMQ Q to pass into OpenAS2 can then be installed in the **receiver** package and the module to push files into the RabbitMQ Q can be placed in the **sender** package.

igwtech:
Hi Chris, 
   The DirectoryPollingModule its just loading outbound files from the filesystem (aka, storage) therefore I think its kinda of a stretch consider it as a "receiver" module. Based on the same rationale, can the MessageFileModule (storage package) be considered a "sender" module since it sends received messages to the filesystem?

   I separated the DirectoryPollingModule into its own package because its implementation is fundamentally different than the other 2 modules on the same package, AS2ReceiverModule and AS2MDNReceiverModule (Functional grouping).
  In both cases the modules inherit from NetModule which deal with handling connections with the network. The DirectoryPollingModule on the other hand inherits from MessageBuilderModule which constructs an AS2 message based on properties of the file and configuration settings. The only common ancestor between both is the BaseReceiverModule which is basically an empty class we could do away with without any mayor implication.

   Maybe "producer" was the wrong nomenclature to use but, I thought that the package reorganization was making the code easier to maintain and extend. 
   Seems like it was a subjective appreciation that depends on the point of view of each developer.

   Bottom line is that I don't want to break backwards compatibility without a proper justified reason.

uhurusurfa:
@igwtech I agree that you could argue it both ways. The directory polling module is both  - it receives files into OpenAS2 and passes them to the AS2Sender. 
Similarly, the MessageFileModule is "sending" files out of OpenAS2 but is receiing files from the AS2 receiver.
That is how it was when I inherited the project and whilst I understand your desirte to restructure, it is cosmetic and makes no difference to the application.

If you want to restructure and make the effort to write an upgrade script, I am open to suggestions.

PR #171, opened 2019-11-11
Minor Method access for plugins

igwtech:
Changes to simplify creating plugins for OpenAS2.
Making Session and Server more extensible.
After the discussion on the PR: CodeRefactor, I decided there was no good enough reason to introduce the refactoring changes, therefore I drop them in favor for minimal needed code changes.

### DIFF
--- a/Server/src/main/java/org/openas2/BaseSession.java
+++ b/Server/src/main/java/org/openas2/BaseSession.java
@@ -55,7 +55,7 @@ public abstract class BaseSession implements Session {
      * @param comp        component to register
      * @see Component
      */
-    void setComponent(String componentID, Component comp)
+    public void setComponent(String componentID, Component comp)
     {
         Map<String, Component> objects = getComponents();
         objects.put(componentID, comp);
--- a/Server/src/main/java/org/openas2/app/OpenAS2Server.java
+++ b/Server/src/main/java/org/openas2/app/OpenAS2Server.java
@@ -55,7 +55,7 @@ public class OpenAS2Server {
                 .run(args);
     }
 
-    private void start() throws Exception
+    protected void start() throws Exception
     {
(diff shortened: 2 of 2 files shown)


uhurusurfa:
@igwtech  @VboxNick 
Javier
If you want to do a major refactor then there is a change I have been wanting to make for a while and would be a breaking change but I think make the cofig a lot simpler for non-tech people.
Instead of having the class specied for each module, I want to use simple names for each of the modules that are easy to understand and have a properties file that maps the module name to the class that implements it.
So instead of:
`   <processor classname="org.openas2.processor.DefaultProcessor"
              pendingMDN="$properties.storageBaseDir$/pendingMDN3"
              pendingMDNinfo="$properties.storageBaseDir$/pendinginfoMDN3">
      <module enabled="$properties.module.AS2SenderModule.enabled$"
              classname="org.openas2.processor.sender.AS2SenderModule"
              retries="5"/>
      <module enabled="$properties.module.MDNSenderModule.enabled$"
              classname="org.openas2.processor.sender.MDNSenderModule"`
It would be:
   <processor name="ModuleProcessor"
              pendingMDN="$properties.storageBaseDir$/pendingMDN3"
              pendingMDNinfo="$properties.storageBaseDir$/pendinginfoMDN3">
      <module enabled="$properties.module.AS2SenderModule.enabled$"
              name="AS2Sender"
              retries="5"/>
      <module enabled="$properties.module.MDNSenderModule.enabled$"
             name="MDNSender"

And the properties file would contain:
ModuleProcessor=org.openas2.processor.DefaultProcessor
AS2Sender=org.openas2.processor.sender.AS2SenderModule
MDNSender=org.openas2.processor.sender.MDNSenderModule

That way it is still entirely pluggable without code changes but makes the config less verbose.
Additionally, any code refactoring can be done independently of the config and only require changing the properties file.

Issue #88, opened 2018-06-04
Improvement: Create new message handler for inbound/outbound messages to support RabbitMQ/others


justanu:
This is a new feature not a bug.
I would like to create a new handler for both inbound and outbound messages:
- outbound messages would be retrieved from RabbitMQ/AMQP instead of file system (or any other messaging system);
- inbound messages will be sent to a dedicated queue in RabbitMQ/AMPQ instead of file system (or any other messaging system).
There will be two queues created, one for inbound, another for outbound. The MQ message will contain sender and receiver (eg AS2 identifier) at the header level.
I would need in config.xml to be able to specify the implementation handler for both inbound and outbound message.
I am willing to work on it once I get the workspace setup.

jsmucr:
Good that you mention this. I've already developed modules that communicate with Apache ActiveMQ Artemis like this. The problem is that OpenAS2 isn't quite ready for this, so the solution is a bit hacky. But we're planning to go productive with this. One day, soon.

uhurusurfa:
@jsmucr  Please feel free to suggest how OpenAS2 can be adapted to make it easier to create a message queue module

@justanu You can follow the suggestions on how to add the inbound and outbound handlers as discussed here: https://sourceforge.net/p/openas2/discussion/265567/thread/75b5def5/
Essentially, for inbound messages implement a class that replaces org.openas2.processor.storage.MessageFileModule for passing received messages into a queue. Declare the module in the same way as the MessageFileModule is declared .

Similarly for outbound messages implement a class that replaces org.openas2.processor.receiver.AS2DirectoryPollingModule cfor receiving messages to be sent via AS2 from the message queue

You should be able to copy the 2 classes lkisted above and then replace the relevant code in the class with code that reads/writes message queues.

jsmucr:
@uhurusurfa Sure. I didn't mean to be rude. I appreciate the software very much and will gladly help with its development (however my boss allows me to :-)).

This is an example of my code (Groovy). There's a nasty thing I had to use to be able to create an `AS2Message` instance with partnership details preset (since there's no input file path to extract these things from):


```
import cz.aimtec.mq.MapUtils
import org.apache.activemq.artemis.api.core.client.ClientMessage
import org.openas2.OpenAS2Exception
import org.openas2.Session
import org.openas2.message.AS2Message
import org.openas2.message.Message

class MessageSourceModule extends ClouEDIMQSourceModule {

[...]

PR #175, opened 2019-11-15
Improvement: Create new message handler for inbound/outbound messages to support RabbitMQ/others #88

igwtech:
This plugin allows OpenAS2 to interconnect with a Message Queue Broker (i.e RabbitMQ) for message exchange and message tracking events.


Plugins-MQ/src/main/java/com/greicodex/openas2/plugins/mq/MessageBuilderModule.java
```diff
@@ -0,0 +1,167 @@
...
+import org.openas2.params.InvalidParameterException;
+import org.openas2.params.MessageParameters;
+import org.openas2.params.ParameterParser;
+import org.openas2.params.RandomParameters;
+import org.openas2.partner.Partnership;
+import org.openas2.processor.receiver.BaseReceiverModule;
+import org.openas2.util.AS2Util;
+
+public abstract class MessageBuilderModule extends BaseReceiverModule {
+
```
uhurusurfa: Why are you re-implementing the MessageBuilderModule?
If there is something that you need in the existing one then lets talk about enhancing it rather than duplicating all the code.

igwtech:
The original uses the filename to get the partnership info and other
metadata.
  MQ messages have the headers for that purpose. That's was basically the
difference.

J

uhurusurfa:
Ok - lets make the retrieval of that info into a method in MessageBuilderModule then it can be overridden by modules that need something diffferenmt if necessary or simply pass additional parameters in the AS2Message context object to.
Break MessageBuilder down into smaller methods that can be overridden wherever necessary.

uhurusurfa:
Stale PR.

PR #186, opened 2020-06-28
Decreased noise in logs

VboxNick:
If a monitoring system invokes the health check each 30s, OpenAS will
generate useless 5760 log entries. Those log entries are noise can
can overwhelm the log analytics systems. The log level for those
two entries has been decreased to DEBUG. Logging for the real data
transfers as well as for the errors still present.


VboxNick:
> 👍

thanks.

PR #194, opened 2020-08-21
Sanitizing payload filename

citizenTwice:
For review: proposed mitigation of path traversal vulnerability, as per separate discussion vie email with developer Christopher Broderick.


citizenTwice:
Hi Javier @igwtech 

Yes, for 'own use', I always stick to that approach and restrict the name to  [ A-Z 0-9 -_. ] only ,which is enough as far as I'm concerned, esp for EDI. But I'm not so sure about OpenAS2 as it is meant to be generally deployable/usable anywhere in the world.  I think we can trust the JRE/JDK here: for path traversal protection I think Path.getFileName is good enough as well as InvalidPathException for catching any illegal characters for a given platform, given how wildly the requirements differ between the various OS, it is unlikely that we would do a better job than the JDK developers if we try to reimplement those checks...
Thanks

igwtech:
Hi LuigiG,
   You have a very good point there. I stand corrected.

J

uhurusurfa:
@citizenTwice - happy with that

citizenTwice:
> @citizenTwice - happy with that

will get coding and add a new commit to this PR

citizenTwice:
@uhurusurfa @VboxNick @igwtech 

Added new commits, please review.

Updates:

The filename sent by partner is now passed to IOUtil::getSafeFilename,
new function that works as follows:

1 - known, bad characters (e.g. 0x0d, 0xa, 0x00) are removed automatically
2 - if the platform-specific nio.file.Path  still rejects the cleaned name 
    InvalidMessageException is thrown
3 - the name is checked for embedded directory paths    
4 - if any are found
    InvalidMessageException is thrown
5 - otherwise the (cleaned) filename is returned to the caller    

I've also tried to add a few unit tests. This turned out to be a bit tricky for me because rejection of some filenames is platform-dependent and the tests need to be deterministic. I've opted to let the tests run only on the platforms where I was myself able to directly verify the test results myself, that's the three main desktop/server OS's (Mac,Linux,Windows) ...but still, I am unsure as whether better solutions exist.

And sorry for the delay following-up on this.

uhurusurfa:
Awesome - thank you for the contribution.

adioss:
@uhurusurfa  sorry to bother you with that but, shouldn't a CVE being created associated to this issue?
Thanks a lot,
Adrien

uhurusurfa:
@adioss It was not a vulnerability

adioss:
@uhurusurfa thanks a lot for the feedback. In fact, to be 100% transparent, I'm an AppSec engineer and Snyk reports this as a security issue. 

- Please check https://security.snyk.io/vuln/SNYK-JAVA-NETSFOPENAS2-608851 I just asked them "why no CVE?" 
- if I may, now I'm wondering: according to the PoC provided in the ticket, why don't you consider this as a security issue?

Adrien

uhurusurfa:
@adioss - although that getSafeFielname() method was added to parse file names, there was already a filename check done in IOUtil.cleanFilename() that is called further down line when the received file is actually persisted in the storage module handler and it is still called to this day so there was never a security vulnerability in OpenAS2. 
What is claimed in Snyk is true that a  filename could be contrived that in theory could do what the reporter  "citizenTwice" claims but "citizenTwice" did not actually test that theory at all because if that claimant had, they would have realised that it was not and is not possible in OpenAS2.

uhurusurfa:
@citizenTwice - Thanks for the clarification. I am surprised and will cross check that the exploit does work without the getSafeFilename() invocation.
@adioss  - If true then I agree it should have been a CVE

PR #208, opened 2021-01-14
fix http timeout handling

jonrios:


uhurusurfa (changes requested):
Thanks for the PR.
Please revert the pom.xml files to HEAD - I will manage the version bump.

uhurusurfa (changes requested):
Actually, that change is not backwards compatible for people who are already using socket timeoutt so cannot accept it.
Leave socket time out and revove the read timeout since it overrides the connect timeout

uhurusurfa (changes requested):
Please run these commands then commit and push and I will be happy to merge:
git checkout 128a567617 Server/pom.xml
git checkout 128a567617 Remote/pom.xml
git checkout 128a567617 Bundle/pom.xml

jonrios:
Hi @uhurusurfa ,

I apologize for the delay.

As you said I reverted the socket timeout removal and removed the read timeout instead. I also added the socket timeout to the HttpSenderModule to make it configurable for sending modules

uhurusurfa (approved):
👍  Thanks for the contribution

Issue #127, opened 2018-10-03
Document content_type setting


uhurusurfa:

jsmucr:
Hello. I'm dealing with this issue too. It would be very easy to be solved if there was an option to use variables (file name parts) within the `mimetype` parameter. Is that possible or shall I try to create a pull request to implement this functionality?

PR #209, opened 2021-01-22
Document content_type setting #127

yuri1969:
This is a simple attempt of describing the Content-Type usage.

I've added a dedicated section 7.5 and modified no. 6 of section 14.


uhurusurfa (approved):
Thank you for contributing to the documentation - much appreciated.

PR #229, opened 2021-08-16
Update Dockerfile and start-container.sh

LeoWare:
Dockerfile:
1. Removed the trailing / in the environment variables. Doing this made the rest of the file look more readable. Dockerfile also accepts variable references in the form ${variable}. This helps when joining variables and strings.
2. Added the destination directories to the COPY commands. All the files were being dumped into the $OPENAS2_HOME directory.
3. Created config directory

start-container.sh:
1. Removed 'cd' command. There is no config_template directory inside the config_template directory.

I would also like to point out that if $OPENAS2_BASE and $OPENAS2_HOME are different, start-container.sh will fail to copy the files and OpenAS2 will complain about no config.xml.

#226

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,15 +17,15 @@ RUN cd /usr/src/openas2/Runtime/bin && \
 
 
 FROM openjdk:11-jre-slim 
-ENV OPENAS2_BASE=/opt/openas2/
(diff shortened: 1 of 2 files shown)


uhurusurfa (approved):
👍  Thanks for the contribution

PR #243, opened 2021-11-08
Support for disable sign time verification

claudiodegio:
Add the properties configuration di disable the certificate expiration time with 

<?xml version="1.0" encoding="UTF-8"?>
<openas2>
  <properties
       ...
       as2_sign_allow_expired_certificate ="true"
       />

Is configured as2_sign_allow_expired_certificate to true, the message or mdn It's allowed to pass if certificate was expired. 
It's log the certificate information in log with warning level like

2021-11-05 17:23:55.163 WARNING BCCryptoHelper: OUTDATED CERTIFICATE signer info: 
    Digest Alg OID: 1.3.14.3.2.26
    Encrypt Alg OID: 1.2.840.113549.1.1.1
    Signer Version: 1
    Content Digest: [-61, -38, -90, -119, -40, 5, 40, 5, -18, -99, -12, 36, 25, 90, 94, 61, 56, 48, 80, 68]
    Content Type: 1.2.840.113549.1.7.1
    SID: C=IT,ST=scaduto,L=scaduto,O=scaduto,OU=scaduto,CN=scaduto

NOTA: the change is needed for ours customer we have many certificate expired


Server/src/main/java/org/openas2/lib/helper/BCCryptoHelper.java
```diff
@@ -396,10 +386,25 @@ public MimeBodyPart verifySignature(MimeBodyPart part, Certificate cert) throws
...
+                    logSignerInfo("Verified signature for signer info", signer, part, x509Cert);
+                    return signedPart.getContent();
+                }
+            } catch (CMSVerifierCertificateNotValidException ex) {
+                String as2SignIgnoreTimeIssue = Properties.getProperty("as2_sign_allow_expired_certificate", "false");
+
+                if ("true".equalsIgnoreCase(as2SignIgnoreTimeIssue)) {
+                    // if flag is enabled log only issue
+                    logSignerWarn("OUTDATED CERTIFICATE signer info", signer, part, x509Cert);
+                    return signedPart.getContent();
```
uhurusurfa, line 402: The signing must still be verified even if the date stamp on the certificate is ignored so use this code to verify the signature inside the catch block ignoring the certificate date if the CMSVerifierCertificateNotValidException exception is thrown:
```
JcaSimpleSignerInfoVerifierBuilder verifierBuilder = new JcaSimpleSignerInfoVerifierBuilder();
signer.verify(verifierBuilder.build(x509Cert));
```

uhurusurfa (changes requested):
Please see the requested change.

Server/src/main/java/org/openas2/lib/helper/BCCryptoHelper.java
claudiodegio: Sorry i dont know very well the implementation the verify check certificare ad signing time ? 
To solve the code reported check certificate but not the exporation ? 

JcaSimpleSignerInfoVerifierBuilder verifierBuilder = new JcaSimpleSignerInfoVerifierBuilder();
signer.verify(verifierBuilder.build(x509Cert));

uhurusurfa:
@claudiodegio  
I have not tested this solution at all but try the following below. It will still validate the signature but will not care about the dates on the certificate:
1. Add this code to a new file  Server/src/main/java/org/openas2/lib/helper/SignerInfoIgnoringExpiredCertificate.java:
```
package org.openas2.lib.helper;

import java.lang.reflect.InvocationTargetException;
import java.lang.reflect.Method;

import org.bouncycastle.cms.CMSException;
import org.bouncycastle.cms.SignerInformation;
import org.bouncycastle.cms.SignerInformationVerifier;

public class SignerInfoIgnoringExpiredCertificate extends SignerInformation{

    protected SignerInfoIgnoringExpiredCertificate(SignerInformation baseInfo) {
        super(baseInfo);
    }

    /**
     * Verify that the given verifier can successfully verify the signature on
     * this SignerInformation object.
     *
     * @param verifier a suitably configured SignerInformationVerifier.
     * @return true if the signer information is verified, false otherwise.
     * @throws org.bouncycastle.cms.CMSVerifierCertificateNotValidException if the provider has an associated certificate and the certificate is not valid at the time given as the SignerInfo's signing time.
     * @throws org.bouncycastle.cms.CMSException if the verifier is unable to create a ContentVerifiers or DigestCalculators.
     */
    public boolean verify(SignerInformationVerifier verifier)
        throws CMSException {
        Method m = null;
        try {
            m = getClass().getSuperclass().getDeclaredMethod("doVerify", new Class<?>[]{});
        } catch (NoSuchMethodException | SecurityException e) {
            // TODO Auto-generated catch block
            e.printStackTrace();
        }
        if (m != null) {
            m.setAccessible(true);
            try {
                return (boolean) m.invoke(this, verifier);
            } catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
                // TODO Auto-generated catch block
                e.printStackTrace();
            }
        }
        return false;
    }

}
```

2. In Server/src/main/java/org/openas2/lib/helper/BCCryptoHelper.java  in the verifySignature method, change this code:
```
                if (signer.verify(signerInfoVerifier)) {
                    logSignerInfo("Verified signature for signer info", signer, part, x509Cert);
                    return signedPart.getContent();
                }
```
... to this:
```
            try {
                if (signer.verify(signerInfoVerifier)) {
                    logSignerInfo("Verified signature for signer info", signer, part, x509Cert);
                    return signedPart.getContent();
                }
            } catch (CMSVerifierCertificateNotValidException e) {
                signer = new SignerInfoIgnoringExpiredCertificate(signer);
                signer.verify(signerInfoVerifier);
            }
```

claudiodegio:
I have changed my code based on suggestion

uhurusurfa (changes requested):
Just a slight logic change requested.

Server/src/main/java/org/openas2/lib/helper/BCCryptoHelper.java
```diff
@@ -396,10 +386,31 @@ public MimeBodyPart verifySignature(MimeBodyPart part, Certificate cert) throws
...
+                    // if flag is enabled log only issue
+                    if (signer.verify(signerInfoVerifier)) {
+                        logSignerWarn("Verified signature for signer info EXCLUDING certificate date verification (OUTDATED CERTIFICATE)", signer, part, x509Cert);
+                        return signedPart.getContent();
+                    }
+                }
+
+            } catch (CMSVerifierCertificateNotValidException ex) {
+                throw ex;
             }
```
uhurusurfa, line 411: Can you change the check so that it only occuirs inside the catch block. That way it is not necessary to change the configuration whe the certificate is updated to have valid dates:

```
            try {
                String as2SignIgnoreTimeIssue = Properties.getProperty("as2_sign_allow_expired_certificate", "false");
                // normal check
                if (signer.verify(signerInfoVerifier)) {
                    logSignerInfo("Verified signature for signer info", signer, part, x509Cert);
                    return signedPart.getContent();
                }
            } catch (CMSVerifierCertificateNotValidException ex) {
                if ("true".equalsIgnoreCase(as2SignIgnoreTimeIssue)) {
                    signer = new SignerInfoIgnoringExpiredCertificate(signer);
                    // if flag is enabled log only issue
                    if (signer.verify(signerInfoVerifier)) {
                        logSignerWarn("Verified signature for signer info EXCLUDING certificate date verification (OUTDATED CERTIFICATE)", signer, part, x509Cert);
                        return signedPart.getContent();
                    }
                }
                } else {
                    throw ex;
                }
            }
```

uhurusurfa (changes requested):

Server/src/main/java/org/openas2/lib/helper/BCCryptoHelper.java
claudiodegio: I have changed the code as requested

claudiodegio:
Thank you

uhurusurfa (approved):
👍

PR #245, opened 2021-11-25
Add no-console rule to WebUI

kek-Sec:
no-console rule is needed to avoid getting errors on build when console.log is found within the src.


uhurusurfa (approved):
👍 Thanks for the contribution

PR #252, opened 2022-01-07
Remove double quotes from as2id on sync mdns

brandonjoyce:
Not sure if opening a PR like this is helpful, but this is the change I needed to make to resolve an issue I was having with synchronous MDN processing.  It appears to be similar to code changes I looked at:

here: https://github.com/OpenAS2/OpenAs2App/pull/93 
and here: https://github.com/OpenAS2/OpenAs2App/pull/92

I was getting a PartnershipNotFound exception when processing an MDN response for a partner who is using a space in their AS2 identifier.  

This change seems to fix the issue.  With some guidance I'm willing to write some tests/clean this up if we think this is something worth pulling into the main project.


uhurusurfa (approved):
👍  Thanks for contributing.

PR #257, opened 2022-02-02
Handle InputStream when reading MDN text

### LINKED ISSUE #253: MDN Response logging shows as Java object instead of the message string
I'm sending some messages using OpenAS2 and receiving the corresponding MDNs with success but when I look into the stored received MDN files (and the mdn_response field in the db), instead of the text received in the MDN, it shows something like:

...
DISPOSITION: automatic-action/MDN-sent-automatically; processed
Text:
java.io.ByteArrayInputStream@76182f20

and it shows the same in the db mdn_response field:

    java.io.ByteArrayInputStream@76182f20

pkraeutli:
Fixes OpenAS2/OpenAs2App#253

### DIFF
--- a/Server/src/main/java/org/openas2/util/AS2Util.java
+++ b/Server/src/main/java/org/openas2/util/AS2Util.java
@@ -1,5 +1,6 @@
 package org.openas2.util;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.openas2.ComponentNotFoundException;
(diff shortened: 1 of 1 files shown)


Server/src/main/java/org/openas2/util/AS2Util.java
```diff
@@ -147,6 +134,16 @@ public static void parseMDN(Message msg, X509Certificate receiver) throws OpenAS
         }
     }
 
+    private static String getMimeBodyPartText(MimeBodyPart part) throws MessagingException, IOException {
+        Object content = part.getContent();
+
+        if (InputStream.class.isAssignableFrom(content.getClass())) {
+            return IOUtils.toString((InputStream) content, StandardCharsets.UTF_8);
```
pkraeutli, line 141: Not sure how to get the actual charset ...

uhurusurfa: After line 119 of parseMDN method add this:
```
                Charset charset = StandardCharsets.UTF_8;
                String charsetFromContentType = reportType.getParameter("charset");
                if (charsetFromContentType != null) {
                	try {
						charset = Charset.forName(charsetFromContentType);
					} catch (Exception e) {
						// Just warn and allow the default to go thru
						logger.warn("Received charset string that cannot be parsed: " + charsetFromContentType + msg.getLogMsgID());
					}
                }
```
Then pass `charset` into the getMimeBodyPartText method.

uhurusurfa (changes requested):
Thanks for the contribution - I have added  comment for determining the character set

uhurusurfa (approved):
👍 Looks good. Thank you for the contribution to the code base.

PR #265, opened 2022-03-08
Fix for h2 migration from 1.4.200 to 2.1.210

kcs-dushyanttankariya:
**Issue:** Failed to persist a tracking event: The write format 1 is smaller than the supported format 2 [2.1.210/5]
**Root Cause:** H2 Database library is upgraded from 1.4.200 to 2.1.210.

**Solution:** 
1. Build Image and run the container.
2. Removed config/DB/openas2.* files.
3. Create new DB file using h2_create_new_db.sh
4. Copy new generated files and replaced with existing code files available under config/DB directory.
5. Updated db_ddl.sql file by replacing IDENTITY to AUTO_INCREMENT keyword on create table SQL.


uhurusurfa (approved):
Thanks for sorting this out.

PR #266, opened 2022-03-08
Avoid IOException when missing Content-Length

brandonjoyce:
Curious if this makes sense to others as I think it makes sense for my team.  This exception happens when someone just opens the as2 url in a browser (happens a lot during setups). The exception causes more log/alerting noise than we desire as we monitor for exceptions being logged pretty closely.  It appears that we're already returning enough information for the user to know what the issue is, and there is also information being logged besides the exception details.

### DIFF
--- a/Server/src/main/java/org/openas2/util/HTTPUtil.java
+++ b/Server/src/main/java/org/openas2/util/HTTPUtil.java
@@ -257,9 +257,11 @@ public class HTTPUtil {
                 return null;
             } else {
                 HTTPUtil.sendHTTPResponse(outStream, HttpURLConnection.HTTP_LENGTH_REQUIRED, null);
-                Log logger = LogFactory.getLog(HTTPUtil.class.getSimpleName());
-                logger.error("Inbound HTTP request does not provide means to determine data length: " + request.get(0) + " " + request.get(1) + "\n\tHeaders: " + printHeaders(msg.getHeaders().getAllHeaders(), "==", ";;"));
-                throw new IOException("Content-Length missing and no \"Transfer-Encoding\" header found to determine how to read message body.");
+                if ("true".equals(Properties.getProperty(Properties.LOG_INVALID_HTTP_REQUEST, "true"))) {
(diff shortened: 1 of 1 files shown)


Server/src/main/java/org/openas2/util/HTTPUtil.java
```diff
@@ -259,7 +259,7 @@ public static String getHTTPResponseMessage(int responseCode) {
                 HTTPUtil.sendHTTPResponse(outStream, HttpURLConnection.HTTP_LENGTH_REQUIRED, null);
                 Log logger = LogFactory.getLog(HTTPUtil.class.getSimpleName());
                 logger.error("Inbound HTTP request does not provide means to determine data length: " + request.get(0) + " " + request.get(1) + "\n\tHeaders: " + printHeaders(msg.getHeaders().getAllHeaders(), "==", ";;"));
-                throw new IOException("Content-Length missing and no \"Transfer-Encoding\" header found to determine how to read message body.");
```
uhurusurfa, line 262: Removing the exception that is thrown malkes sense.
However, reviewing the code it seems the error message is not accurate as there is no guarantee that is why there was no data.
Given that for some users of OpenAS@ it seems the error can occur quite frequently, it is probably a good idea to make the erro log into a warning and make the warning suppressable using this:
```
if ("true".equals(Properties.getProperty(Properties.LOG_INVALID_HTTP_REQUEST, "true"))) {
    logger.warning("The request either contained no data or has issues with the Transfer-Encoding or Content-Length: : " + request.get(0) + " " + request.get(1) + "\n\tHeaders: " + printHeaders(msg.getHeaders().getAllHeaders(), "==", ";;"));
}
```
If you make this additional change I will merge it.

uhurusurfa (changes requested):

Server/src/main/java/org/openas2/util/HTTPUtil.java
brandonjoyce: Perfect, I'll get to this very soon.

uhurusurfa (approved):
👍 Thanks for the contribution

PR #272, opened 2022-04-26
Honour http.nonProxyHosts property

pkraeutli:
Currently, the `http.nonProxyHosts` property is ignored when determining if a proxy should be used.

This PR adds this, although using the `DefaultProxySelector` for the whole proxy logic would probably be better. I did it that way in order to touch as little as possible of the existing code.

### DIFF
--- a/Server/src/main/java/org/openas2/util/HTTPUtil.java
+++ b/Server/src/main/java/org/openas2/util/HTTPUtil.java
@@ -1,47 +1,5 @@
 package org.openas2.util;
 
-import java.io.BufferedInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.DataInputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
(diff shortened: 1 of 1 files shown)


uhurusurfa (commented):
Thanks for the contribution.
It should be taken care of by the HTTPClient lib is you add these lines to the setProxyConfig method:
```
        SystemDefaultRoutePlanner routePlanner = new SystemDefaultRoutePlanner(null);
        builder.setRoutePlanner(routePlanner);
```

uhurusurfa (commented):
Thanks for the contribution.
It should be taken care of by the HTTPClient lib is you add these lines to the setProxyConfig method:
```
        SystemDefaultRoutePlanner routePlanner = new SystemDefaultRoutePlanner(null);
        builder.setRoutePlanner(routePlanner);
```

pkraeutli:
@uhurusurfa thanks for the hint!

PR #270, opened 2022-04-18
FIX:  Unable to access rest api #268 

igwtech:
This pull request fixes issues in the RESTAPI and updates dependencies on WebUI

PR #274, opened 2022-05-23
Fix: JacksonFeature not registered

fridolin-koch:
It looks like that this PR https://github.com/OpenAS2/OpenAs2App/pull/270 broke the JSON encoding. This broke the WebUI. I readded `JacksonFeature.class`, assuming it was removed unintentionally. 

Before:

Screenshot from 2022-05-23 10-44-21

With the fix in this PR:

Screenshot from 2022-05-23 10-42-38


uhurusurfa (approved):
👍  - Thanks for contributing

PR #284, opened 2022-07-14
set CLASSPATH in start-openas2.bat

Chri-s:
The `set` command was missing before the CLASSPATH so that it generated a "command not found" message.


uhurusurfa (approved):
👍  thanks for the fix

PR #288, opened 2022-08-07
Update usage message in import_public_cert.bat

Chri-s:
The angle brackets were not escaped in the usage message. So the batch script stopped with an error message when it should display the usage message.

Also the batch filename will be used in the usage message (that is the ``%~nx0``).


uhurusurfa (approved):
👍  thanks for the PR - much appreciated

PR #292, opened 2022-08-17
Fix smtpauth property and set to default smtp port

Chri-s:
The property email.**smtpauth** was referenced as "$properties.email.**auth**$". This prevented the server to start if email logging was enabled.

Also the default value of the property email.smtpport is changed to the default smtp port (25).


uhurusurfa (approved):
👍 thanks for the fix

PR #293, opened 2022-08-18
Fix high cpu consumption in RestCommandProcessor

### LINKED ISSUE #283: High cpu load when enabling the rest api
I extracted the newest release 3.2.1 and enabled the rest api by setting `restapi.command.processor.enabled="true"`. When starting OpenAS2, the java process will continously consume one cpu core with 100%. Nothing else of the configuration was changed from the default.
If I disable the rest api and restart OpenAS2, almost no cpu is consumed.

I tested this in 2 virtual machines:

Windows Server 2019 with
java version "17.0.3.1" 2022-04-22 LTS
Java(TM) SE Runtime Environment (build 17.0.3.1+2-LTS-6)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.3.1+2-LTS-6, mixed mode, sharing)

Ubuntu 20.04 with
openjdk version "11.0.15" 2022-04-19
OpenJDK Runtime Environment (build 11.0.15+10-Ubuntu-0ubuntu0.20.04.1)
OpenJDK 64-Bit Server VM (build 11.0.15+10-Ubuntu-0ubuntu0.20.04.1, mixed mode, sharing)

Both environments show the same behavior.

In the Windows Server 2019 VM I also tested the versions 3.2.0, 3.1.0, 3.0.0 and in the Ubuntu VM the version 2.14.0.

To test that it is not the fault of HyperV, I also checked version 3.2.1 on the Windows 10 host with Java 17.0.3.1.

The behavior exists in all tested versions.

Chri-s:
RestCommandProcessor.processCommand() did nothing and BaseCommandProcessor.schedule called processCommand in an endless loop, resulting in a thread which continuously processed an endless loop.

By overriding schedule there is no loop any more.

Closes #283.

### DIFF
--- a/Server/src/main/java/org/openas2/cmd/processor/RestCommandProcessor.java
+++ b/Server/src/main/java/org/openas2/cmd/processor/RestCommandProcessor.java
@@ -42,8 +42,13 @@ public class RestCommandProcessor extends BaseCommandProcessor {
     private HttpServer server;
 
     @Override
+    public void schedule(ScheduledExecutorService executor) throws OpenAS2Exception {
+        // Do nothing. Everything is set up in the method init.
+    } 
+
+    @Override
     public void processCommand() throws Exception {
-        //throw new UnsupportedOperationException("Commands received by HTTP Server thread");
+        // This method is never called because schedule is overriden.
     }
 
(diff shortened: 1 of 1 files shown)


uhurusurfa (approved):
🚀  - thanks for the fix - much appreciated

PR #315, opened 2023-04-04
Fix various issues with the WebUI

brianpeiris:
The WebUI does not build correctly without some fixes. This PR fixes those issues and allows you to run the WebUI out of the box.

- Pinned the node.js to version 16 in Dockerfile_WebUI, since the libraries used do not support newer versions of node.js
- Cleaned up docker instructions in README.md, and added instructions for Dockerfile_WebUI
- Enabled the REST API by default, so that it's easier to use the WebUI
- Removed references to vuetify, since vuetify was removed in f2740226
- Fixed eslint parser configuration. This was previously causing errors during the build process
- Set the WebUI login server to "http" by default, instead of "https"
- Updated yarn.lock with the dependency changes

### DIFF
--- a/Server/src/config/config.xml
+++ b/Server/src/config/config.xml
@@ -108,7 +108,7 @@
                         response_format="xml"/>
       <commandProcessor classname="org.openas2.cmd.processor.RestCommandProcessor" 
                         enabled="$properties.restapi.command.processor.enabled$"
-                        baseuri="http://0.0.0.0:8443"
+                        baseuri="http://localhost:8443"
                         ssl_protocol="TLS"
                         ssl_keystore="$properties.ssl_keystore$"
                         ssl_keystore_password="$properties.ssl_keystore_password$"
--- a/Server/src/config/config.xml
+++ b/Server/src/config/config.xml
@@ -11,7 +11,7 @@
                email.logger.enabled="false"
                console.command.processor.enabled="true"
                socket.command.processor.enabled="false"
-               restapi.command.processor.enabled="false"
+               restapi.command.processor.enabled="true"
                module.AS2SenderModule.enabled="true"
                module.AS2SenderModule.readtimeout="60000"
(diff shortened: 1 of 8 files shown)


Server/src/config/config.xml
```diff
@@ -108,7 +108,7 @@
                         response_format="xml"/>
       <commandProcessor classname="org.openas2.cmd.processor.RestCommandProcessor" 
                         enabled="$properties.restapi.command.processor.enabled$"
-                        baseuri="http://localhost:8443"
```
GreicodexJM, line 111: Hi Brian,
   This configuration is used not only on Docker Images, but also on Standalone servers. Therefore for security reasons I lean towards having a more secure by default configuration limiting the exposure of the API to localhost and SSL.

   An alternative solution for running the WebUI as a docker image could be to either provide more documentation related to the configuration changes needed or to include a sample config.xml file.

brianpeiris: Makes since. I've pushed a new commit to change it back to localhost, and I've adjusted the WebUI docker instructions instead.

uhurusurfa (approved):
⭐ - thanks for the contribution getting this addon to OpenAS2 usable again.

Server/src/config/config.xml
Chetan-Khopade: > Makes sense. I've pushed a new commit to change it back to localhost, and I've adjusted the WebUI docker instructions instead.

Thanks for the update,
Now, I've setup new openas2 server version 3.6.0 with webUI and created some Partners configuration and connection for testing but i am not able to see the data location of the Partners, connections, certificates, send and received folder path inside of the amazon linux.
Could you please help me to where i can found the path of the data folders inside in Linux machine.

brianpeiris: Sorry, I can't help with that. I'd recommend you create a new discussion to get help from others. https://github.com/OpenAS2/OpenAs2App/discussions

PR #321, opened 2023-06-02
Added Docker compose and updated README with instructions

ThorTL67:
Extends the Docker instructions to include Docker Compose.


uhurusurfa (approved):
👍 Thanks for the contribution

PR #332, opened 2023-07-30
Webui fix not persisting changes to partnerships

uhurusurfa:
Fix the H@ database secuirity issue:

- Upgrade the inital databases to support the latest release
- Provide an upgrade script to upgrade existing databases

Swithc toi JUNIT 5 test library

Various fixes to the REST command interface and underlying command processing:

- Fix persistence of partnership adding and deleting
- Fix deleting partner and partnership nodes


uhurusurfa:
Ah - ok. The referenced PR is highly unlikely to be the source of your issue. That only rewrites the partnerships.xml when it receives a an update or add command from the UI app so unless your contqiner is somehow sending a command art startup, the source of the problem must be elsewhere.
Off the top of my head I cannot provide any advice but will give it some thought and come back to you.

uhurusurfa:
See this for resolution: #346

PR #361, opened 2024-03-07
Allow config.xml parameters to be overridden with environment variables

### LINKED ISSUE #360: WebUI network error
Hello, so i ran the docker compose file in the directory, and i keep getting the network error when trying to login with the provided credentials, 
im not sure whats the issue, 
attached below are the logs for both the webUi and the server itself, it seems the request is not passing from the UI to the server as there is nothing related to the request in the server log.

image

image

igwtech:
The default configuration listens on 127.0.0.1 internally in the docker container of OpenAS2 therefore the docker-compose port exposure doesn't route into it.

This pull request allows the config.xml parameters to be overridden with environment variables on the docker-compose.yml

### DIFF
--- a/Dockerfile_WebUI
+++ b/Dockerfile_WebUI
@@ -4,6 +4,8 @@ COPY ./WebUI /usr/src/webui
 WORKDIR /usr/src/webui
 RUN yarn install
 RUN npx browserslist@latest --update-db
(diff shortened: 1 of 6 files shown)


uhurusurfa (approved):
👍

PR #366, opened 2024-04-15
Updating Dockerfile to more recent base images

PapaToemmsn:
The used openjdk:11 package in the Dockerfile was last built 2 years ago and contains several vulnerabilities (see https://hub.docker.com/layers/library/openjdk/11-jre-slim/images/sha256-884c08d0f406a81ae1b5786932abaf399c335b997da7eea6a30cc51529220b66?context=explore).

Thus I have used eclipse-temurin:11.0.22_7-jre-jammy as a base which is actively supported.

I also switched to nginx_stable-alpine  for the WebUI dockerfile as it is smaller.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,10 @@
-FROM openjdk:11 AS builder
+FROM eclipse-temurin:11.0.22_7-jdk-jammy AS builder
 COPY . /usr/src/openas2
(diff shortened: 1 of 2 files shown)


uhurusurfa (approved):
Thanks for the contribution

PR #378, opened 2024-05-28
Display JDBC driver name during startup

apeiris:
Display JDBC driver name


Server/src/main/java/org/openas2/processor/msgtracking/DbTrackingModule.java
```diff
@@ -278,23 +274,23 @@ public HashMap<String, String> showMessage(String msg_id) {
...
         try {
             if (useEmbeddedDB) {
                 conn = dbHandler.getConnection();
             } else {
                 conn = DriverManager.getConnection(jdbcConnectString, dbUser, dbPwd);
             }
+            CallableStatement cs = conn.prepareCall("{CALL ftGetMessagesByDateRange(?,?)}");
+            cs.setDate(1, Date.valueOf(map.get("startDate")));
+            cs.setDate(2,Date.valueOf(map.get("endDate")));
+            ResultSet rs = cs.executeQuery();
```
uhurusurfa, line 288: This is your own custom callable.
It will not be added to OpenAS2 since the goal is to remain generic.
I assume the statement that is currently used does not work on SqlServer because it does not support the TIMESTAMP operator so instead change the statement to this and it will work on all SQL-92 and newer standards compliant database servers:
```
            ResultSet rs = s.executeQuery("SELECT " + FIELDS.MSG_ID + ",STATE,STATUS,CREATE_DT FROM " + tableName
                    + " WHERE CREATE_DT BETWEEN CAST('" + map.get("startDate").toString()
                    + " 00:00:00' as TIMESTAMP) AND CAST('" + map.get("endDate").toString()
                    + " 23:59:59'" as TIMESTAMP));
```

Server/src/config/config.xml
uhurusurfa, line 1: There is no functional change to this file. Revert to HEAD for this PR.

Server/pom.xml
```diff
@@ -298,5 +298,11 @@
 		    <groupId>javax.annotation</groupId>
 		    <artifactId>javax.annotation-api</artifactId>
 		</dependency>
+        <!-- https://mvnrepository.com/artifact/com.microsoft.sqlserver/mssql-jdbc -->
+        <dependency>
+            <groupId>com.microsoft.sqlserver</groupId>
+            <artifactId>mssql-jdbc</artifactId>
+            <version>12.6.2.jre11</version>
+        </dependency>
```
uhurusurfa, line 306: Please remove this import as it loads a library into the installler package that most people will never use and is not part of the default OpenAS2 functionality.
Anyone wishing to use packages other than the standard packages provided by OpenAS2 should create their own wrapper installer otherwise the OpenAS2 install package will become bloated unnecessarily.

uhurusurfa (changes requested):
Thanks for the PR.
I have made comments below that will need to be made before this PR can be approved.

PR #384, opened 2024-06-20
Log JDBC driver name on server startup

apeiris:
Log JDBC driver name on server startup


uhurusurfa (approved):
👍 Thanks for the enhancement

PR #385, opened 2024-07-10
Add API endpoint to retrieve configuration properties

apeiris:
Added a new API endpoint to retrieve configuration properties from the config.xml file.

### Changes
- Created an endpoint at `/api/config/properties` to return configuration properties as JSON.

### Testing
- Verified the endpoint returns the correct JSON structure as expected.
-Added unit testclass ApiResourceTest

### DIFF
--- a/Server/src/main/java/org/openas2/cmd/processor/restapi/ApiResource.java
+++ b/Server/src/main/java/org/openas2/cmd/processor/restapi/ApiResource.java
@@ -7,9 +7,13 @@ package org.openas2.cmd.processor.restapi;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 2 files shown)


uhurusurfa (approved):
Looks good.

uhurusurfa:
It is failing because the jakarta package does not support Java 8.

[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] /home/runner/work/OpenAs2App/OpenAs2App/Server/src/test/java/org/openas2/ApiResourceTest.java:[4,26] cannot access jakarta.ws.rs.core.Response
  bad class file: /home/runner/.m2/repository/jakarta/ws/rs/jakarta.ws.rs-api/3.1.0/jakarta.ws.rs-api-3.1.0.jar(jakarta/ws/rs/core/Response.class)
    class file has wrong version 55.0, should be 52.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
[INFO] 1 error

You will have to exclude the resource test from the Java 8 run .... look in the .github/workflows/actions.yml for how it has been done for other API tests for Java 8 only

apeiris:
Close #385 and resubmit without running this specific test

PR #394, opened 2024-10-02
Resend file cleanup error

uhurusurfa:
Prevent call to file cleanup when message goes into resend queue by providing and explicit bollena that is set by the reender module on the Message object.


uhurusurfa (commented):
👍

PR #401, opened 2024-11-04
Improvement: Additional way to handle self-signed certificates if necessary

szabarna:
Description

This pull request introduces an enhancement to HTTPUtil.java in OpenAS2 to improve handling of self-signed certificates. Previously, the only option for trusting self-signed certificates was the TrustSelfSignedCN system property, which is not ideal in most cases.

Problem Statement

In our setup, we must accept self-signed certificates from trading partners when sending messages to them. Additionally, in a Kubernetes environment, relying on system properties is cumbersome and inflexible. Configuring trusted self-signed certificates with TrustSelfSignedCN would require a Docker image rebuild whenever certificates change, limiting the ability to update certificates independently of the application image.

Solution

This update allows OpenAS2 to dynamically load trusted self-signed certificates from a keystore file specified by environment variables:

    SSL_KEYSTORE_PATH: Path to the keystore containing trusted certificates.
    SSL_KEYSTORE_PASSWORD: Password for the keystore.

With these environment variables set, HTTPUtil.java will bypass the need for TrustSelfSignedCN and use the specified keystore for SSL validation.

Changes Made

    Modified HTTPUtil.java to check if SSL_KEYSTORE_PATH and SSL_KEYSTORE_PASSWORD environment variables are set.
    If the environment variables are present, OpenAS2 loads the specified keystore and validates self-signed certificates against it.
    If the keystore contains a matching certificate, hostname verification is skipped, allowing the self-signed certificate to be trusted without rebuilding the image.

Benefits

    Kubernetes Compatibility: Enables dynamic, environment-based configuration without requiring image rebuilds.
    Improved Maintainability: Certificates can be updated by updating the keystore file mounted in the container, without modifying code or system properties.

Testing

This change has been tested in a Kubernetes environment to confirm that OpenAS2 dynamically reads the keystore and correctly validates self-signed certificates against it.

Summary

I just wanted to ask if there will be an update in the future that enhances the way self-signed certificates are handled in a more flexible manner, as the current TrustSelfSignedCN option was not sufficient for our needs.

In this pull request, I’ve introduced a solution that leverages environment variables to define the keystore path and password, making it possible to dynamically load trusted certificates without relying on system properties or requiring image rebuilds in Kubernetes environments.

Is this approach a good starting point, or is there a planned update that might handle self-signed certificates in an even more flexible way?

### DIFF
--- a/Server/src/main/java/org/openas2/util/HTTPUtil.java
+++ b/Server/src/main/java/org/openas2/util/HTTPUtil.java
@@ -48,6 +48,7 @@ import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 
 
 /**
@@ -73,7 +74,9 @@ public class HTTPUtil {
 
     public static final String SSL_KEYSTORE_PATH_ENV = "SSL_KEYSTORE_PATH";
     public static final String SSL_KEYSTORE_PASSWORD_ENV = "SSL_KEYSTORE_PASSWORD";
-    private static final ThreadLocal<Set<String>> trustedFingerprints = ThreadLocal.withInitial(HashSet::new);
+    private static Set<String> cachedFingerprints = ConcurrentHashMap.newKeySet();
+    private static KeyStore cachedCustomKeyStore = null;
+
     private static final Logger LOG = LoggerFactory.getLogger(AS2ReceiverHandler.class);
 
     public abstract static class Method {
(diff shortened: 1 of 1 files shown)


uhurusurfa:
Thanks for the PR. I have had a quick scan through and the proposal looks better than the current implementation for self signed certificates.
I have a few changes I would like to suggest and will try to get around to it in the next few days.

uhurusurfa (approved):
👍 Thanks for the contribution. After going through all the changes necessary for the ultimate solution I decided to merge yours then make the changes myself for how this will be deployed into the nexrt release.
It will changed to use the PKCS12CertificateFactory module so that it is automaitcally monitored and reloaded and can be managed through an API.

uhurusurfa:
@szabarna - the latest PR is a further enhancement o what you staretd. It is fully documented in the OpenAS2HowTo now
#404

Issue #415, opened 2025-03-10
http.nonProxyHosts does not work without http.proxyUser, with quotes and more related issues


markieww:
The fix https://github.com/OpenAS2/OpenAs2App/pull/272 to support the `http.nonProxyHosts` does not work. 

We use a configuration with an exception for hosts which should not be routed via a proxy.

For example: `-Dhttp.proxyHost=proxy -Dhttp.proxyPort=80 -Dhttps.proxyHost=proxy -Dhttps.proxyPort=80 -Dhttp.nonProxyHosts="localhost|.local|externalhost"`

But when sending an AS2-message to https://externalhost:8443/content/foo.bar.AS2Server (configured in partnerships.xml) via OpenAS2 4.0.0 the proxy is still used.

markieww:
Please have a look at my mergerequest https://github.com/OpenAS2/OpenAs2App/pull/416

uhurusurfa:
This is now released in v4.1.1

PR #416, opened 2025-03-10
http.nonProxyHosts does not work without http.proxyUser, with quotes …

markieww:
…and more related issues #415

### DIFF
--- a/Server/src/main/java/org/openas2/util/HTTPUtil.java
+++ b/Server/src/main/java/org/openas2/util/HTTPUtil.java
@@ -716,9 +716,17 @@ public class HTTPUtil {
             throw new OpenAS2Exception("Missing PROXY port since Proxy host is set");
         }
         int port = Integer.parseInt(proxyPort);
-        HttpHost proxy = new HttpHost(proxyHost, port);
 
-        rcBuilder.setProxy(proxy);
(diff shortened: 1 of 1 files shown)


uhurusurfa (changes requested):
In principal this looks like a good solution.

The one catch in it is that the proxy configuration can be done either asa system property or as an AS2 property in the properties section of the config.xml or as a properties file provided to AS2 at startup.

So to use this solution you will need to push the proxy properties into the **System** properties if they are sourced from the **Properties** object.

I suggest that at the entry to the  method, make the first check to the **System** property with fallback to the Properties and if found in **Properties** object then search for all possible proxy properties in Properties object and set them into the System property.

markieww:
@uhurusurfa Here is the updated PR with the following changes

- respect proxy configuration from properties
Note the behaviour-change of using the user/password!
BEFORE: Use http.proxyUser/http.proxyPassword for https.proxyHost (<-- I guess that was an issue and not intentional)
NOW: Use https.proxyUser/https.proxyPassword for https.proxyHost
- introduce http.nonProxyHosts for configuration via properties
- Cleanup - Remove unused methods in HTTPUtil

Please review.

uhurusurfa (approved):
Looks great - thanks for the contribution. 🚀

PR #432, opened 2025-03-27
Set header value for Content-Transfer-Encoding for every request

xxxyyyz:
regardless of property use_old_mime_deserialise_method


uhurusurfa (approved):
Looks good - probably need to create a unit test to test for this hack which supports non-conformant AS2 systems that do not set the header on the mime body part but that will have to be another day.

PR #446, opened 2025-06-03
Dynamic content type fix

szabarna:
Problem

Our AS2 integration partner relies on the Content-Type header to determine how to process received files. To support this, we wanted to use OpenAS2’s dynamic content-type lookup feature, which maps file extensions to MIME types using a properties file.

However, despite configuring the feature correctly (both at the system and partnership level), the Content-Type was never resolved dynamically at runtime.
Root Cause

After adding debug logs and tracing execution, we discovered that while the Partnership objects were configured correctly (e.g., useDynamicContentTypeMapping=true and mapping files loaded), the Partnership instance used during message processing was a copy of the original.

The method Partnership#copy(...) did not copy over the internal fields related to dynamic lookup, namely:

this.useDynamicContentTypeLookup = partnership.useDynamicContentTypeLookup;
this.contentTypeFromFileExtensionMap = partnership.contentTypeFromFileExtensionMap;
this.overrideContentTypeFromFileExtensionMap = partnership.overrideContentTypeFromFileExtensionMap;

As a result, even if the original Partnership had dynamic lookup enabled and properly loaded mappings, the copied instance behaved as if the feature was disabled.
Fix

We added the three missing fields to the copy(...) method of the Partnership class:

this.useDynamicContentTypeLookup = partnership.useDynamicContentTypeLookup;
this.contentTypeFromFileExtensionMap = partnership.contentTypeFromFileExtensionMap;
this.overrideContentTypeFromFileExtensionMap = partnership.overrideContentTypeFromFileExtensionMap;

This ensures that the copied instance retains all necessary dynamic lookup configuration.
Why the Tests Didn't Catch It

The DynamicContentTypeTest test suite uses the original Partnership object directly — it never triggers a copy(). As such, dynamic lookup worked in the test scenario but failed in production, where the copied instance is used during message handling (e.g., in MessageBuilderModule).
Result

With this fix, dynamic content-type lookup now functions as intended in real deployments. The correct Content-Type is determined based on file extension mappings defined in system-level or partnership-specific configuration files.


uhurusurfa (approved):
🌟 Awesome - thanks for debugging the issue and submitting a PR - very much appreciated.

PR #451, opened 2025-06-24
Bug: Mismatched build argument in WebUI Dockerfile prevents API URL configuration

HamzaDevz:
Hi there,

I've found a small bug in the `Dockerfile_WebUI` that prevents configuring the backend API URL at build time. It seems to be caused by a simple variable name mismatch.

The Vue.js application code expects an environment variable named `VUE_APP_RESTAPI_URL`:

```javascript
// From the Vue.js source
server: process.env.VUE_APP_RESTAPI_URL || 'http://127.0.0.1:8443/api'
```

However, the `Dockerfile_WebUI` declares and sets a build argument with a different name, `VUE_APP_API_URL` (it's missing "REST"):

```dockerfile
# From Dockerfile_WebUI
ARG VUE_APP_API_URL
ENV VUE_APP_API_URL=${VUE_APP_API_URL:-http://localhost:8080}
```

Because the names don't match, any value passed as a `build-arg` is ignored by the application, and it always uses its internal fallback URL (`http://127.0.0.1:8443/api`).

A simple fix would be to align the variable name in `Dockerfile_WebUI` to match what the application code expects.

Hope this helps\! Thanks for the great work on the project.


uhurusurfa (approved):
Thanks for the fix - much appreciated. 💯

PR #466, opened 2025-08-25
Feature/api secure endpoints

apeiris:
Remove redundant comment regarding XPath validation


apeiris:
Security & CodeQL fixes applied:

Whitelisted XPath expressions to prevent injection.

Wrapped XML results in a safe root element.

Disabled DOCTYPE parsing to prevent XXE.

Enforced SSL/TLS for non-local requests.

All CodeQL/security issues resolved.

Server/src/main/java/org/openas2/cmd/processor/restapi/ApiResource.java
```diff
@@ -220,6 +247,139 @@ public Response headCommand(@PathParam("param") String command) {
...
+
+            XPathExpression xPathExpr = XPathFactory.newInstance().newXPath().compile(xpathExpression);
+            nodeList = (NodeList) xPathExpr.evaluate(document, XPathConstants.NODESET);
+
+        } catch (Exception ex) {
+            LoggerFactory.getLogger(ApiResource.class.getName()).error("Error parsing XML file: " + xmlFileName, ex);
+            // return null on error
+        }
+        return nodeList;
+    }
```
uhurusurfa, line 381: This appears to be a redundant method.
What is the reason this is here?

Server/src/main/java/org/openas2/cmd/processor/restapi/ApiResource.java
```diff
@@ -220,6 +247,139 @@ public Response headCommand(@PathParam("param") String command) {
...
+
+        if (xpathExpression == null || !allowedXPaths.contains(xpathExpression)) {
+            return Response.status(Response.Status.BAD_REQUEST)
+                    .entity("{\"error\":\"Invalid or disallowed XPath expression\"}")
+                    .type(MediaType.APPLICATION_JSON)
+                    .build();
+        }
+
+        Session session = getProcessor().getSession();
+        String filePath = session.getBaseDirectory() + "/" + filename;
```
uhurusurfa, line 308: Since there is no validation of the file path this could retrieve any XML file on the server and is a security risk.
Is there a reason why you need to pass a filename in at all since you are only interested in the partnships file?

Server/src/main/java/org/openas2/cmd/processor/restapi/ApiResource.java
```diff
@@ -220,6 +247,139 @@ public Response headCommand(@PathParam("param") String command) {
...
+    public Response getPropertyList(@Context Request request) {
+        if (!request.isSecure() && !isLocalhost(request)) {
+            return Response.status(Response.Status.FORBIDDEN)
+                    .entity("{\"error\":\"SSL/TLS required\"}")
+                    .type(MediaType.APPLICATION_JSON)
+                    .build();
+        }
+        Map<String, String> result = new HashMap<>();
+        try {
+            result = (Map<String, String>) Properties.getProperties();
```
uhurusurfa, line 263: Note that his returns the parsed properties not the p[roperties as set in the properties file.
If any properties use property references within their value then those references will have been replaced.
Can you provide a use case for returning the cached properties?

uhurusurfa (changes requested):
Please can you provide more insight into why you are retirieving XML from disk using OpenAS2 to return to a service that is running on the same server?
Would it not be better to just read the file directly in the other service?

Server/src/main/java/org/openas2/cmd/processor/restapi/ApiResource.java
uhurusurfa: This method is not called from anywhere so I am unclear why it is here unless you have other code that is not included in this PR.

Server/src/main/java/org/openas2/cmd/processor/restapi/ApiResource.java
apeiris: Yes, intended usage is primarily for administration and debugging. Regarding the passwords, would redacting properties before returning the properties address the concern? I can implement that to ensure sensitive information is not exposed

uhurusurfa: Yes - if you remove any property that has the word "password" in it's property name or property value that should work.
These are the currently known properties that have password information:
```
               as2_keystore_password="testas2"
               ssl_trust_keystore_password="testas2"
               ssl_keystore_password="testas2"
               restapi.command.processor.password="pWd"
               module.HealthCheckModule.keystore_password="$properties.ssl_keystore_password$"
               msg_tracking.tcp_server_password="openas2"
```

PR #468, opened 2025-08-29
Feature/api secure enpoints

apeiris:
Add Jakarta WS RS API dependency and enhance ApiResource XML/JSON handling

- Added jakarta.ws.rs-api in pom.xml
- Refactored ApiResource to:
  - Add getPropertyList endpoint with sensitive property redaction
  - Improve getXml endpoint with proper XML parsing, XXE protection, and sensitive attribute removal
  - Implement redactSensitiveAttributes method to remove password/pwd fields from XML nodes

### DIFF
--- a/Server/OPENAS2_LOG_DIR_IS_UNDEFINED/log-20250828.txt
+++ b/Server/OPENAS2_LOG_DIR_IS_UNDEFINED/log-20250828.txt
@@ -1,0 +1,1124 @@
+20:57:24.410 [main] INFO  org.openas2.app.OpenAS2Server -- Retrieving config file...
+20:57:24.865 [main] INFO  org.openas2.app.OpenAS2Server -- Using MANIFEST file:/D:/REPOS/apeiris/OpenAs2App/Server/target/dist/lib/openas2-server-4.6.0.jar!/META-INF/MANIFEST.MF
+20:57:24.915 [main] INFO  org.openas2.XMLSession -- Loading configuration...
+20:57:24.916 [main] INFO  org.openas2.XMLSession -- Loading properties...
+20:57:24.918 [main] DEBUG org.openas2.XMLSession -- Parsing property: app.base.dir : D:\REPOS\apeiris\OpenAs2AppMyConfig
+20:57:24.920 [main] DEBUG org.openas2.XMLSession -- Parsing property: module.AS2MDNReceiverModule.http.enabled : true
+20:57:24.920 [main] DEBUG org.openas2.XMLSession -- Parsing property: email.bodytemplate : D:\REPOS\apeiris\OpenAs2AppMyConfig/emailtemplate.txt
+20:57:24.928 [main] DEBUG org.openas2.XMLSession -- Parsing property: as2_keystore_password : testas2
+20:57:24.928 [main] DEBUG org.openas2.XMLSession -- Parsing property: module.AS2ReceiverModule.http.enabled : true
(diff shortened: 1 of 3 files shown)


Server/src/main/java/org/openas2/cmd/processor/restapi/ApiResource.java
```diff
@@ -220,6 +243,164 @@ public Response headCommand(@PathParam("param") String command) {
...
+                    .build();
+        }
+    }
+
+    @GET
+    @RolesAllowed({"ADMIN"})
+    @Path("/getXml")
+    @Produces({MediaType.APPLICATION_XML,MediaType.APPLICATION_JSON})
+    public Response getXml(@Context Request request,
+                           @QueryParam("filename") String filename,
```
uhurusurfa, line 289: To be clear:
`        I will not support passing in a filename to be retrieved. It is a security risk since the user could pass a path to any file on the server using '../../etc'`

If you are only interested in the partnerships XML file then there is no need to pass in a file name at all.
Additionally, the file names should not need to be known by the user of the UI and the partnerships file could be located anywhere on the server using any user defined name for the file itself so you should be retrieving the file using the call to the partnership factory method.

If you feel there is a need to retrieve more than just the partnerships XML file from the file system then you should pass in an identifier and map the supported set of identifiers to appropriate getters to retrieve the file.

If you want to be able to retrieve the config XML file then please provide a use case as the config.xml is not really intended to be modifed directly but instead controlled by properties set in an external properties file and I do not see what the benefit of retrieving the config XML file would be.

The partnerships file path can be retrieved using this:
`String partnershipsFile = getProcessor().getSession(). getPartnershipFactory().getFilename()`

uhurusurfa (changes requested):

Server/OPENAS2_LOG_DIR_IS_UNDEFINED/log-20250828.txt
uhurusurfa, line 1: Please remove this file

uhurusurfa (commented):
Please remove the log file as well

Server/src/main/java/org/openas2/cmd/processor/restapi/ApiResource.java
apeiris: Thanks for the clarification — I agree with your concern around not exposing filenames directly and avoiding path traversal risks.

One point of clarification: there isn’t currently an existing method like

String partnershipsFile = getProcessor().getSession().getPartnershipFactory().getFilename();


in the codebase.

Given that, would it be acceptable if I implement a REST endpoint that returns the list of partnerships using:

getProcessor().getSession().getPartnershipFactory().getPartnerships();


This way, the endpoint exposes the in-memory partnerships directly via the factory without relying on file names, while keeping to the security and design guidelines you’ve outlined.

uhurusurfa: Zh yes - the getFilename method in the XMLPartnershipFactory is not public so you would have to change the method to public then do this:
`String partnershipsFile = ((XMLPartnershipFactory)getProcessor().getSession().getPartnershipFactory()).getFilename();`

However, I will support accessing the list of partnerships directly.

apeiris: Thanks , I’d prefer accessing the list of partnerships directly. That approach feels cleaner

Issue #485, opened 2025-10-23
Provide an option to disable the quotes for the filename at the content-disposition header


markieww:
When `sendfilename="true"`  is set at module-level of an AS2DirectoryPollingModule in config.xml the filename is sent as header `content-disposition: Attachment; filename="xxxx.csv"`

The target AS2-Server we like to communicate with is quite picky about the quotes around the filename. So I like to introduce an option to toggle the quoting of this filename.

**Proposed behaviour**

The configuration has to take place at partnership-level.
* Include the quotes when `<attribute name="quote_send_file_name" value="true"/>` or no option is given (<- current behaviour is the default for keeping the compatibility) 
   *  results in `content-disposition: Attachment; filename="xxxx.csv"`
* Exclude the quotes when `<attribute name="quote_send_file_name" value="false"/>` 
   * results in `content-disposition: Attachment; filename=xxxx.csv`

I will provide a pull request.

markieww:
Hi @uhurusurfa again, please review the pull request https://github.com/OpenAS2/OpenAs2App/pull/486.

uhurusurfa:
👍  I will take a look this weekend.

PR #486, opened 2025-10-23
Pull Request for #485:  Provide an option to disable the quotes for the filename at the content-disposition header 

markieww:
See #485

Sorry, I was not able to write a test for this change. I played around in org.openas2.app.OpenAS2ServerTest#verifyMessageDelivery but I had no access to the `msgheaders` directory. There was no getter in TestPartner nor TestMessage to access the file with the header information which I liked to check.

So I tested manually and it was successful. (I used a internal config.xml and partnership.xml which I am not allowed to share, but you get the idea how it works)
````
# Partnership without configuration
echo "EMPTY" > /opt/openas2/data/outbox/AAAA/empty.txt
# Partnership with quotes = true
echo "TRUE" >  /opt/openas2/data/outbox/BBBB/true.txt
# Partnership with quotes = false
echo "FALSE" > /opt/openas2/data/outbox/CCCC/false.txt

cat /opt/openas2/data/pendingMDN3/* | grep -a "Attachment"

Content-Disposition: Attachment; filename="empty.txt"
Content-Disposition: Attachment; filename="empty.txt"
Content-Disposition: Attachment; filename="true.txt"
Content-Disposition: Attachment; filename="true.txt"
Content-Disposition: Attachment; filename=false.txt
Content-Disposition: Attachment; filename=false.txt
````

### DIFF
--- a/Server/src/config/partnerships.xml
+++ b/Server/src/config/partnerships.xml
@@ -69,7 +69,9 @@
 		-->
         <!--
         Example for disabling the quoting of the sent filename at partnership-level.
-        Works only in combination with sendfilename="true" at module-level of AS2DirectoryPollingModule in config.xml.
+        Works only in combination with sendfilename="true" using pollerConfigBase.sendfilename="true" in the config.xml OR
+        at partnership-level in the partnerships.xml using &lt;pollerConfig enabled="true" sendfilename="true"/&gt; 
+
         <attribute name="quote_send_file_name" value="false"/>
         -->
     </partnership>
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Server/src/config/partnerships.xml [hunk 1]
```diff
@@ -67,6 +67,11 @@
         <attribute name="split_file_contains_header_row" value="true"/>
         <attribute name="split_file_name_prefix" value="SF"/>
 		-->
+        <!--
+        Example for disabling the quoting of the sent filename at partnership-level.
+        Works only in combination with sendfilename="true" at module-level of AS2DirectoryPollingModule in config.xml.
```
uhurusurfa: Same as earlier comment about how the sendfilename attribute is used.

markieww:
@uhurusurfa : I added a test and updated the documentation. Please have a look

Issue #537, opened 2026-06-14
db_ddl.sql is missing `sent_file_name` column — breaks DbTrackingModule when schema is created manually


ricardopsj:
## Environment
- OpenAS2 v4.8.2 (built from source — no binary published in this release)
- Java 21 (OpenJDK)
- H2 embedded (file mode), `tcp_server_start="false"`
- Debian 13

## Issue
`config/db_ddl.sql` (used to bootstrap the H2 tracking schema) does not include the
`SENT_FILE_NAME` column that is present in the authoritative DDL Utils schema at
`resources/db/openas2-schema.xml`.

When DbTrackingModule tries to insert a tracking record after a successful outbound,
it fails because the column is missing in the actual table.

## Reproduction
1. Configure DbTrackingModule with `tcp_server_start="false"` and embedded H2.
2. Initialize the schema using `db_ddl.sql` via `org.h2.tools.RunScript`
   (since `create_db_table.sh` requires `ant` + TCP 9092, both unavailable in a
   typical hardened deployment).
3. Send any outbound message.
4. DbTrackingModule logs an error trying to write `sent_file_name`.

## Workaround
Add the missing column manually to `db_ddl.sql` before running it:
    ALTER TABLE msg_metadata ADD COLUMN sent_file_name VARCHAR(255);
(or include it in the original CREATE TABLE statement, aligning with
 `resources/db/openas2-schema.xml`).

## Suggested fix
Regenerate `config/db_ddl.sql` from `resources/db/openas2-schema.xml` so the two
stay in sync; or remove `db_ddl.sql` from the distro and document the schema
bootstrap to use `openas2-schema.xml` directly through DDL Utils.

## Side note
`openas2-schema.xml` cannot be used directly with `org.h2.tools.RunScript`
because it's DDL Utils XML, not plain SQL — users without `ant` end up using
`db_ddl.sql`, which is why the missing column matters.

uhurusurfa:
Your fix is in release 4.8.3

PR #538, opened 2026-06-14
Fix db_ddl.sql: align with authoritative openas2-schema.xml

ricardopsj:
## Summary

Aligns `Server/src/config/db_ddl.sql` with the authoritative schema in `Server/src/resources/db/openas2-schema.xml`. The two had drifted: most notably, `db_ddl.sql` was missing the `SENT_FILE_NAME` column that `DbTrackingModule` writes on every outbound, plus several other smaller divergences (column types, defaults).

Fixes the bug described in #537 (created alongside this PR).

## What's in the change

- **`SENT_FILE_NAME VARCHAR(255)`** added immediately after `FILE_NAME` (same position as in the XML).
- **`MSG_ID`, `PRIOR_MSG_ID`, `MDN_ID`, `MDN_RESPONSE`, `STATE_MSG`** promoted from unbounded `VARCHAR` to `LONGVARCHAR`, matching the XML.
- **Column defaults** added so the SQL matches the XML's behavior on fresh installs:
  - `IS_RESEND DEFAULT 'N'`
  - `RESEND_COUNT DEFAULT 0`
  - `CREATE_DT DEFAULT CURRENT_TIMESTAMP`
- **Header comment** added pointing maintainers at `openas2-schema.xml` as the single source of truth, so the two files don't drift again.
- **`Server/src/test/resources/config/db_ddl.sql`** mirrored with the same content to keep the test fixture aligned.

## Why this matters

`create_db_table.sh` (the "official" bootstrap path) needs Apache Ant on the host and TCP 9092 open on the H2 server. Hardened deployments routinely:

- skip ant (not on the runtime image), and
- set `tcp_server_start="false"` in `DbTrackingModule` so H2 stays in embedded/file mode (no exposed 9092).

In those deployments the only viable bootstrap is to apply `db_ddl.sql` directly via `org.h2.tools.RunScript` against the embedded `.mv.db` file. That's the path this PR fixes.

## Backwards compatibility

- Pure additive on the schema (one new column, defaults on existing columns, type widenings). No data migration needed for fresh installs.
- For existing installs that already created the table from the old `db_ddl.sql` and want to upgrade in place:
  ```sql
  ALTER TABLE msg_metadata ADD COLUMN sent_file_name VARCHAR(255);
  ```
  (`DbTrackingModule` inserts will work after this even without restarting.)

## How this was tested

- Bootstrapped a fresh H2 file with the new `db_ddl.sql` using `org.h2.tools.RunScript` on Debian 13 + OpenJDK 21.
- Sent an outbound message; `DbTrackingModule` wrote a complete row including `sent_file_name`.
- Smoke test repeated on a second independent deployment (same OS/JDK stack), same result.

## Notes for reviewers

- The XML schema (`openas2-schema.xml`) remains the source of truth; this PR just brings the convenience SQL up to date with it.
- A follow-up could consider auto-generating `db_ddl.sql` from the XML at build time so they cannot drift again. Out of scope for this PR — happy to do it as a second PR if maintainers prefer.

### DIFF
--- a/Server/src/config/db_ddl.sql
+++ b/Server/src/config/db_ddl.sql
@@ -1,22 +1,25 @@
--- ----------------------------------------------------------------------- 
--- msg_metadata 
--- ----------------------------------------------------------------------- 
+-- -----------------------------------------------------------------------
+-- msg_metadata
+-- -----------------------------------------------------------------------
+-- This DDL must stay aligned with the authoritative schema in
+-- Server/src/resources/db/openas2-schema.xml. If you add/remove a column
+-- here, mirror the change in the XML (and vice versa).
 
 DROP TABLE msg_metadata IF EXISTS;
 
--- ----------------------------------------------------------------------- 
--- msg_metadata 
--- ----------------------------------------------------------------------- 
+-- -----------------------------------------------------------------------
+-- msg_metadata
+-- -----------------------------------------------------------------------
 
 CREATE TABLE msg_metadata
 (
     ID INTEGER NOT NULL AUTO_INCREMENT,
-    MSG_ID VARCHAR NOT NULL,
-    PRIOR_MSG_ID VARCHAR,
-    MDN_ID VARCHAR,
+    MSG_ID LONGVARCHAR NOT NULL,
+    PRIOR_MSG_ID LONGVARCHAR,
+    MDN_ID LONGVARCHAR,
     DIRECTION VARCHAR(25),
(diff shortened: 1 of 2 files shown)


uhurusurfa (approved):
Thanks for the fixes.
The next release for OpenAS2 will be a major release using Hibernate to manage the schema so the use of DDL utils and the XML schema will soon be dropped.

PR #547, opened 2026-06-24
Enhance ViewPartnershipCommand to include pollerConfig

mrconlisk:
Added functionality to extract and append pollerConfig from XML partnerships.


uhurusurfa (approved):
Looks reasonable.

PR #553, opened 2026-07-02
Implement destroyPartnershipPoller method

SohamBhambare:
Added method to destroy and untrack the poller for a deleted partnership, allowing immediate freeing of the outbox directory. This allows us to immediately post after deleting a partnership without getting any poller directory conflicts.


uhurusurfa (approved):
Thanks for the contribution to the project.

PR #554, opened 2026-07-07
Adding from/to parameters to the /messages/list endpoint to filter messages based on date

SohamBhambare:
Adding from/to parameters to the /messages/list endpoint to filter messages based on date.


uhurusurfa (approved):
Nice work. 🚀

PR #556, opened 2026-07-08
Added optional parameters to specify the times for the timeframe between from & to dates.

SohamBhambare:
Extended the /api/messages/list date-range filter (originally added to fix an OOM from unbounded queries) with optional time-of-day precision (fromTime=/toTime=), so callers can bound results to an exact date+time window instead of only whole calendar days — with validation that rejects ambiguous or contradictory inputs rather than silently guessing.


uhurusurfa (approved):
Thanks for the contributions ...  with unit tests 🚀

PR #559, opened 2026-07-17
Add database-backed partnership factory supporting all OpenAS2 databases

JakeHuneau:
Adds org.openas2.partner.DbPartnershipFactory as an alternative to XMLPartnershipFactory, storing partners and partnerships in database tables instead of the partnerships XML file. Works with any JDBC database (embedded H2, MSSql, PostgreSQL, MySQL, Oracle) using the same connection parameters as the DbTrackingModule.

- New tables partner, partner_attribute, partnership and partnership_attribute added to openas2-schema.xml (ddlutils, for all supported platforms) and db_ddl.sql (H2). Attribute categories: attribute, sender/receiver overrides and pollerConfig.
- Partner/partnership add and delete console and REST API commands dispatch on factory type and write to the database, including pollerConfig.* parameters.
- Per-partnership directory pollers are launched from pollerConfig category rows, reusing the pollerConfigBase merge and $partnership.*$ placeholder logic shared via BasePartnershipFactory.
- In-memory state reloads at startup, after every API change, via the refresh command and on an optional fixed interval.
- Tracking module DB handlers made public for reuse by the factory.

Verified with unit tests plus a live server run exercising the REST API (add/view/delete partners and partnerships, poller launch, persistence across restart).


JakeHuneau:
This will likely require a change to the documentation PDF. Is that somewhere that can be edited? Or do you want a write-up of what should be added?

uhurusurfa:
> This will likely require a change to the documentation PDF. Is that somewhere that can be edited? Or do you want a write-up of what should be added?

FYI: There is a major new version (5.X) in the pipeline that moves away from XML based configuration and requires major changes to the documentation. The changes in V5 are not backwards compatible for the most part and includes:
- much more flexible configuration (property based making it much simpler to understand for non-technical users and easier to implement a UI to allow browser based configuration)
- certificate generation and management supported via in Java rather than thru command line scripts
- changes the way the server supports additional software modules separating handlers from worker threads
- replace all existing database related code to use Hibernate as the database interface allowing use of separate databases for message logs and configuration related information if desired (e.g a document DB for message logs and a SQL DB for configuration and user security)
- complete rework of the API to a more modern mapping including the API request execution chain to make it easier to rationalise and support adding new routes through configuration
- limited initial configuration to get the app running
- much simpler startup scripts for both NIX and Windows
- upgrade of core Java code to use the modern additions in Java 17 and up
- user account management for accessing the API with different levels of access
- can be used in any servlet compliant web server


Version 4.X will be maintained for security releases for a short while after the release of v5.

uhurusurfa (approved):
LGTM 👍

PR #561, opened 2026-07-18
Add JmsPollingModule: queue-based outbound sender

JakeHuneau:
Adds a generic AMQP 1.0 JMS consumer (Apache Qpid JMS) as an alternative outbound intake to AS2DirectoryPollingModule. Works against any AMQP 1.0 broker via the broker_url alone.

An external producer drops the file on shared storage and publishes a queue message whose string properties carry the sender/receiver AS2 IDs and the file path. The consumed file is moved into the existing MessageBuilderModule pipeline (identical signing, encryption, MDN and tracking) and ends up in the sent or error directory exactly like the directory poller.

- Reuses processDocument(File) so the original file is moved (not copied) into the pipeline, matching AS2DirectoryPollingModule behaviour.
- Per-message routing is injected via a new overridable getSenderReceiverDefaults hook on MessageBuilderModule (default behaviour for the directory poller is unchanged).
- Transacted JMS sessions: commit on a successful send (sync MDN received, or async MDN send accepted), roll back on failure so the broker redelivers and dead-letters after its max delivery count. Internal file-based resend is disabled per-message so the broker is the sole retry authority.
- File splitting is not supported for queue intake and fails loudly.

Tested with unit tests against an embedded ActiveMQ Artemis broker plus a live OpenAS2 run verifying successful send (file moved to sent dir) and failure (file moved to error dir, message dead-lettered).


uhurusurfa (approved):
👍 Nice work.

PR #562, opened 2026-07-18
Send outbound HTTP Basic auth pre-emptively

JakeHuneau:
Basic auth credentials were only sent in response to a 401 challenge: a "preemptive-auth" attribute was placed on the request context but nothing consumed it, so authentication never happened pre-emptively. Partners that expect credentials on the first request (without issuing a challenge) were never authenticated.

Populate an AuthCache with the target host so the credentials are sent on the first request, replacing the dead context attribute. Adds a test using a non-challenging mock server to confirm the Authorization header is present on the first request.


uhurusurfa (approved):
🎸

PR #564, opened 2026-07-18
Sanitize partner Message-ID used as inbound payload filename fallback

JakeHuneau:
When a received message has no content-disposition filename and no configured filename fallback property, the payload filename fell back to the raw Message-ID header, which is partner-controlled. Unlike the content-disposition path (which runs through IOUtil.getSafeFilename), this fallback was unsanitized, allowing a crafted Message-ID to influence the stored file path (path traversal).

Run the fallback through IOUtil.getSafeFilename, and use a safe generated name if the Message-ID cannot be made safe. The sanitization primitive is already covered by FilenameSafetyTest.


uhurusurfa:
Remove the as2_certs.p12 from your PR's as it is just wasting time with conflict resolution. It is not necessary to commit the new one generated by the build script every time.

JakeHuneau:
> Remove the as2_certs.p12 from your PR's as it is just wasting time with conflict resolution. It is not necessary to commit the new one generated by the build script every time.

Done

uhurusurfa (approved):
🚀

PR #565, opened 2026-07-18
Fix discarded %home% substitution in DbTrackingModule JDBC connect string

JakeHuneau:
String.replace returns a new string; its result was discarded, so a literal %home% in the jdbc_connect_string was never substituted with the config base directory. Assign the result. No effect on the default config (where %home% is already resolved upstream), but fixes it for a connect string that uses %home% directly.


uhurusurfa (approved):
👍

PR #566, opened 2026-07-18
Fix DeletePartnerCommand silently reporting success on failed delete

JakeHuneau:
When removal of the partner node from the partnerships XML failed, the error CommandResult was constructed but never returned, so the command fell through to TYPE_OK. Because the partner is already removed from the in-memory map before the XML delete, this left memory and the XML out of sync while telling the caller the delete succeeded. Return the error result.


uhurusurfa (approved):
👍

PR #567, opened 2026-07-18
Fix SQL injection via message ID in DbTrackingModule.persist

JakeHuneau:
The existence-check SELECT concatenated the message ID directly into the SQL, and for inbound messages that value is the partner-controlled Message-ID header. A crafted header could inject SQL (e.g. "x' OR '1'='1" made a new message match an existing row, corrupting tracking) or break the query outright, and persist() swallows the resulting exception so the tracking record was silently lost.

Parameterise the existence-check SELECT and escape the message ID in the UPDATE WHERE clause using the existing formatField() helper (the same escaping the SET values already use). Adds a test proving crafted IDs are stored/matched as opaque literals rather than interpreted as SQL.


uhurusurfa (approved):
👍

PR #568, opened 2026-07-19
Add unit tests for DirectoryResenderModule

JakeHuneau:
DirectoryResenderModule had ~13% coverage. Adds tests for the resend queue mechanics: queueing a message (handle), the time-based scheduling (isTimeToSend and scanDirectory), and resend execution (processFile) - verifying the retry count is incremented and the file is deleted on success, and archived to the error directory on failure. The processor is mocked so the resend is observed without a running server. Raises DirectoryResenderModule coverage to ~71%.

The failure test documents an existing double-archive defect in the error path: processFile archives the file, then calls file.delete() on the already-moved file and archives a second time, which throws because the file is gone.


uhurusurfa (approved):
🚀

PR #574, opened 2026-07-22
Add API endpoint to look up the stored MDN file path by filename

JakeHuneau:
Records where each MDN is stored so it can be retrieved later without reconstructing the path from a template.

- New mdn_file_path column on msg_metadata (db_ddl.sql and openas2-schema.xml).
- MDNFileModule records the stored MDN path on the message; BaseMsgTrackingModule writes it to the tracking record. The MDN is now stored before the final state is tracked (send side) so the path is captured, and the receive side re-tracks after storing.
- New messages/mdnpath command and REST endpoint (GET /api/messages/mdnpath/<filename>) returning the MDN file path for a message matched by its received or sent payload filename. The command uses the tracking DB directly so it does not require the optional message factory component.

Verified with unit tests plus a live run: a real send/MDN round-trip records the path and the endpoint returns it, pointing at the MDN file on disk.

**NOTE: This is a breaking change since it requires the DB alter**


uhurusurfa (approved):
👍

PR #592, opened 2026-08-16
Don't fail on duplicate partnership

JakeHuneau:
Currently, if a second partnership is added which matches an existing one already, it will cause a total failure which will cause a PollerModule to stop working. On duplicates, alert on duplicates but use the first one defined so keep functionality.


uhurusurfa (approved):
👍  thanks for the improvement.

PR #596, opened 2026-08-31
Add global switch to disable partnership directory pollers

JakeHuneau:
Adds an optional "enabled" attribute to the pollerConfigBase element in config.xml that suppresses every directory poller declared in a partnership.

Previously the only way to stop an instance sending was to turn off the pollerConfig element on each outbound partnership, which forced deployments that split receiving and sending across separate instances to maintain a second copy of the partnerships file. The base pollerConfig could not be used for this because the partnership attributes are merged over the base ones, so a partnership's enabled="true" always won.

The switch is read once when the base poller config is loaded and gates both partnership factories before the poller component is constructed, so a disabled instance also stops creating the outbox and error directories for partnerships it will never send from. It defaults to enabled when the attribute is absent so existing configurations are unaffected, and it does not apply to poller modules declared directly as a module element in config.xml. Partnerships still load normally so inbound messages continue to resolve.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -19,6 +19,13 @@ This is a minor enhancement release.
    404 both when the message ID is unknown and when the MDN it recorded is no longer on disk, with different messages for
    the two cases. Requires the DB tracking module and a storage module that records the MDN file path (see the
    mdn_file_path column added in 4.9.0).
+2. Add a "partnershipPollersDisabled" property that globally suppresses every directory poller declared in a partnership.
+   It defaults to "false" so existing configurations are unaffected. Setting it to "true" stops the instance sending from
+   any outbox directory while it continues to receive normally, which allows a single partnerships file (or database) to be
+   shared between a sending instance and a receiving only instance instead of maintaining a second copy with every outbound
+   pollerConfig turned off. It can be set in the config.xml properties element or in the OpenAS2 properties file, so a
+   containerised deployment can share one config.xml and decide per instance. Poller modules declared directly as a
+   "module" element in config.xml are not affected by it.
 
 
 Version 4.10.0 - 2026-08-17
--- a/Server/src/config/config.xml
+++ b/Server/src/config/config.xml
@@ -75,6 +75,7 @@
                msg_tracking.tcp_server_port="9092"
                msg_tracking.tcp_server_password="openas2"
                reject_unsigned_messages="false"
+               partnershipPollersDisabled="false"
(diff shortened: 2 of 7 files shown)


uhurusurfa (changes requested):
This implementation changes the paradigm on the use of properties that provide a default to attributes on an element and you can then change the default for the element that then propagates to all instances of that element. 

I see the point of what you are doing but I would prefer that you use a property name that is explicitly a global flag rather than a default for a defined element attribute which then gets used as a global flag.

So if you changed the name from:
`pollerConfigBase.enabled="true"`
to something like:
`partnershipPollersDisabled="false"`

Then reference it as a property rather than an attribute on an element in the relevant method:
    public boolean isPartnershipPollingDisabled() {
        return "true".equalsIgnoreCase(Properties.getProperty("partnershipPollersDisabled", "false");
    }

That would make it immediately obvious what the flag is intended to do.

JakeHuneau:
> This implementation changes the paradigm on the use of properties that provide a default to attributes on an element and you can then change the default for the element that then propagates to all instances of that element. 
> 
> 
> 
> I see the point of what you are doing but I would prefer that you use a property name that is explicitly a global flag rather than a default for a defined element attribute which then gets used as a global flag.
> 
> 
> 
> So if you changed the name from:
> 
> `pollerConfigBase.enabled="true"`
> 
> to something like:
> 
> `partnershipPollersDisabled="false"`
> 
> 
> 
> Then reference it as a property rather than an attribute on an element in the relevant method:
> 
>     public boolean isPartnershipPollingDisabled() {
> 
>         return "true".equalsIgnoreCase(Properties.getProperty("partnershipPollersDisabled", "false");
> 
>     }
> 
> 
> 
> That would make it immediately obvious what the flag is intended to do.

Changed it to a property that should be more obvious as to what's happening. I put it on BasePartnershipFactory instead of Session since it's the common parent to both factories that use it.

One note on liking this approach is that it can set in the openas2.properties to overwrite config.xml.

uhurusurfa:
Resolve the conflicts and I will approve and merge this

uhurusurfa:
Also, klet me know if you have other inbound enhancments otherwise I will make a new release once this is merged.

JakeHuneau:
> Also, klet me know if you have other inbound enhancments otherwise I will make a new release once this is merged.

Merge resolved. I don't have any other inbound enhancements planned at this time.

uhurusurfa (approved):
Nice addition for clustered deployments. 🚀

uhurusurfa:
Release 4.11.0 just out.

PR #597, opened 2026-08-31
Add API endpoint to download the stored MDN for a message

JakeHuneau:
The MDN file path has been recorded in the tracking database since 4.9.0 but there was no way to retrieve the MDN itself: messages/mdnpath only returns the path, which is of no use to a caller that is not on the server's file system.

Adds GET /api/messages/mdn/<msgId>, which resolves the AS2 message ID against the tracking database and streams the recorded file back as an attachment. The message ID is the unique key of the tracking table, so at most one row can match.

The caller only ever supplies a message ID, never a path, so the endpoint cannot be used to read arbitrary files: the path comes from the database and is written there by the MDN storage module. The content originates from a trading partner, so it is served as an attachment with nosniff set rather than as anything a browser will render inline, and the file name is reduced to safe characters before it goes into the Content-Disposition header. The endpoint requires the ADMIN role like every other API endpoint.

A tracking record outlives the file it points at, so an MDN that has been archived or cleaned up returns 404 with a message that distinguishes it from a message ID that matched nothing at all.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -10,6 +10,17 @@ The zip file contains a PDF document (OpenAS2HowTo.pdf) providing information on
 ## NOTE: Testing covers Java 11 to 21.
 ##       Java 8 is NO LONGER SUPPORTED.
 
+Version UNRELEASED
+===========================
+
+This is a minor enhancement release.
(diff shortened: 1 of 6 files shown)


uhurusurfa (approved):
Nice work. 👍

PR #601, opened 2026-09-02
Resend tracking fixes

mrconlisk:
Fixes to include file_name in dbtracking on resends, correctly identifying is_resend in dbtracking, and only increment resend_count once on message resend.


uhurusurfa (approved):
Thanks for the code enhancements and fix 👍

PR #603, opened 2026-09-08
Add tool to migrate a partnerships XML file into the database

JakeHuneau:
`DbPartnershipFactory` can read partnerships from the database instead of the XML file, but there was no way to get an existing configuration in there short of retyping it. This adds a migration tool.

```
bin/upgrade/migrate_partnerships_to_db.sh <partnerships.xml> <jdbc_url> <db_user> <db_password>
bin/upgrade/migrate_partnerships_to_db.sh <partnerships.xml> --dry-run
```

`org.openas2.upgrades.MigratePartnershipsToDb` plus a wrapper script, following the existing `MigratePollingModuleConfig` / `config_transform.sh` convention.

## Mapping

Derived from what each factory does on load, so the migrated database is equivalent rather than approximately right:

| XML | Database |
|---|---|
| `<partner name="X" a="1"/>` | `partner(NAME='X')` plus `partner_attribute` rows, excluding `name` since `loadPartners` synthesises it from the row |
| `<sender name="A"/>` | `partnership.SENDER_PARTNER_ID` resolves to A's row |
| `<sender name="A" x509_alias="z"/>` | plus `partnership_attribute(CATEGORY='sender', ...)` for the override |
| `<attribute name= value=/>` | `CATEGORY='attribute'` |
| `<pollerConfig .../>` | `CATEGORY='pollerConfig'` |

## Notable decisions

- **Values are stored unresolved.** Both factories run `attributeEnhancer` when they load, so resolving `$properties.storageBaseDir$` at migration time would freeze a value meant to stay dynamic. The shipped file's `subject=File $attributes.filename$ sent from $sender.name$...` has to survive verbatim; there is a test for it.
- **One transaction, and the whole file is validated before anything is written**, so a file the schema cannot represent leaves the database untouched.
- **It refuses to run against populated tables**, since that would silently merge two configurations. `--replace` is required to overwrite.
- **`--dry-run`** parses, validates and reports what would be written without connecting to write.

## One case that cannot be migrated

A partnership whose `sender`/`receiver` element carries inline IDs instead of naming a partner is valid in the XML file but unrepresentable in the schema, which requires every partnership to reference a partner row. Rather than inventing a partner, the tool aborts naming the partnership and what to change. `--dry-run` surfaces these before any write.

## Testing

144 tests pass, 11 new. The main one migrates the shipped `partnerships.xml`, loads it back through the real `DbPartnershipFactory`, and asserts it produces the same partnerships `XMLPartnershipFactory` produced from the file, comparing sender IDs, receiver IDs and every attribute, so a dropped attribute or override cannot pass unnoticed. Also verified end to end from the CLI against an H2 file database, including the abort-on-populated and `--replace` paths.

Note that `config/db_ddl.sql` drops `msg_metadata`, so only the partner and partnership section should be run to create the tables. The usage text and wrapper both warn about this.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -9,6 +9,18 @@ The release download file is: OpenAS2Server-4.11.0.zip
 The zip file contains a PDF document (OpenAS2HowTo.pdf) providing information on installing and using the application.
 ## NOTE: Testing covers Java 17 to 21.
 
+Version UNRELEASED
+===========================
+
+This is a minor enhancement release.
+1. Add a migration tool that loads a partnerships XML file into the partner and partnership database tables read by
+   DbPartnershipFactory, so an existing XML configuration can be moved to the database store without being retyped:
+   bin/upgrade/migrate_partnerships_to_db.sh <partnerships.xml> <jdbc_url> <db_user> <db_password>. Sender and receiver
+   overrides and pollerConfig entries are carried across as their respective attribute categories, and values are stored
+   unresolved so placeholders such as $properties.storageBaseDir$ keep working. Pass --dry-run to validate the file and
+   see what would be written without touching the database, or --replace to overwrite partnerships already stored there.
+   The whole migration runs in one transaction, so a file that cannot be represented in the schema leaves it untouched.
+
 Version 4.11.0 - 2026-09-02
 ===========================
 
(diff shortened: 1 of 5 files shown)


uhurusurfa (approved):
LGTM

